PR #4, opened 2018-03-28
Add enabled property (like OrbitControls)

jbyte:
Hello,

this RP adds the **enabled** property to **CameraControls**, so if needed you can disable the controls.


yomotsu:
@jbyte Thanks for your contribution.

It looks good, however the JS file you changed is bundled file with Rollup.js. Could you change [the source file](https://github.com/yomotsu/camera-controls/blob/master/src/camera-controls.js)?

jbyte:
@yomotsu I'm not sure what you mean. I changed the _src/camera-controls.js_ file then ran _npm run release_ and commited everything.

yomotsu:
ahh! Sorry you are right!

yomotsu:
merged! many thanks

yomotsu:
ver 1.2 with `enabled` prop is available on npm.😊

eriksachse:
I was thinking about this problem as well since I have custom camera controls. I just made a new store value called `domhover` that I set to true or false with onMouseEnter and onMouseLeave. Then I disable the camera controls depending on that state.

PR #13, opened 2018-11-27
`saveState()` is now deprecated, using `pushDefaultState()` instead

FMS-Cat:
手始めに、変数名を書き換えてしまいました。
ここから、 `target` , `position` , `zoom` をひとまとめにした構造体的な概念を作って、いろいろな形で使えるようにできればとにらんでいます。

## I did a change that might cause compatibility break

- `saveState()` -> `pushDefaultState()`
  - I don't know the context how `saveState` named, but I'm so confused by the name.
    I think `pushDefaultState` is more proper to this procedure.
  - There still are older one, along with deprecation warning.
    (using `saveState()` on this revision doesn't kill you instantly 😇)

## Other internal variable changes:

- `target0`, `position0`, `zoom0` -> `defaultTarget`, `defaultPosition`, `defaultZoom`
  - I think it's more proper
- `state` (in mouse/touch context) -> `controlState`
  - since it's too ambiguous with other "state" concepts
- `STATE` enums -> `CONTROL_STATE`
  - same reason as above one

## JSON compatibility stuff

I changed the name of `target0`, `position0` as I said above, and these variables are exposed via JSON,
so I needed to add this section in order to support older JSONs:

https://github.com/FMS-Cat/camera-controls/blob/74bca5f50731d9ea514d26104c7b1ad49d41c390/src/camera-controls.js#L672-L676

It also adds version information into JSON and it's required to perform compat stuff,
so **don't forget to raise the version before publish these changes!**
（バージョンまわりのしきたりがわからなかったです🙇）

### DIFF
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -42,9 +42,9 @@ label{
 	<br>
 	<button onclick="cameraControls.moveTo( 3, 5, 2, true )">move to( 3, 5, 2 )</button>
 	<button onclick="cameraControls.fitTo( mesh, true )">fit to the bounding box of the mesh</button>
-	<button onclick="cameraControls.reset( true )">reset</button>
 	<br>
-	<button onclick="cameraControls.saveState()">saveState</button>
+	<button onclick="cameraControls.pushDefaultState()">set default state</button>
+	<button onclick="cameraControls.reset( true )">reset</button>
 	<br>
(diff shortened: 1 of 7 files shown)


FMS-Cat:
Ahhhh I see!!
Yeah I was pretty sure there are some context behind these variable names,
And it's fair enough to not change these names.

I should investigate the other way to do around with target-position states.

PR #15, opened 2018-11-27
We should use Math.exp instead

FMS-Cat:
current method we're doing interpolation at `update()` can do weird behavior under extreme situation (laggy environment like 10FPS, or higher damping factor such as `0.5` or `1.0` or w/e).
We should use `Math.exp` instead here, so it guarantees `target` / `spherical` to be converged to desired location (even if damping factor is `Infinity` )!

It has minor compatibility break (does different behavior under high damping factor) but I recommend to apply this change.

background


yomotsu:
Good point! thanks for your PR.👍

PR #17, opened 2018-11-28
`setLookAt` / `lerpLookAt` now receives number literals instead of `THREE.Vector3`

FMS-Cat:
Currently other API does not depend on THREE's root itself but `setLookAt` and `lerpLookAt` does.
It's no cool! (yes it's definitely my fault, sorry 😖 )

So I made it possible to receive JavaScript `Array` instead of `THREE.Vector3`.

### DIFF
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -542,8 +542,11 @@ export default class CameraControls {
 
 	setLookAt( position, target, enableTransition ) {
 
-		this._targetEnd.copy( target );
-		this._sphericalEnd.setFromVector3( _v3.subVectors( position, target ) );
+		const _position = toVector3( position );
+		const _target = toVector3( target );
+
+		this._targetEnd.copy( _target );
+		this._sphericalEnd.setFromVector3( _v3.subVectors( _position, _target ) );
 		this._sanitizeSphericals();
 
 		if ( ! enableTransition ) {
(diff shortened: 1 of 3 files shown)


src/camera-controls.js
```diff
@@ -542,8 +542,11 @@ export default class CameraControls {
 
 	setLookAt( position, target, enableTransition ) {
```
yomotsu: how about just taking number literals?
```suggestion
	setLookAt( positionX, positionY, positionZ, targetX, targetY, targetZ, enableTransition ) {
```

src/camera-controls.js
```diff
@@ -542,8 +542,11 @@ export default class CameraControls {
...
 
-		this._targetEnd.copy( target );
-		this._sphericalEnd.setFromVector3( _v3.subVectors( position, target ) );
+		const _position = toVector3( position );
```
yomotsu: then we can reuse `_v3` which declared at the beginning 
https://github.com/yomotsu/camera-controls/blob/master/src/camera-controls.js#L21
```suggestion
		const _position = _v3.set( positionX, positionY, positionZ );
```

src/camera-controls.js
```diff
@@ -542,8 +542,11 @@ export default class CameraControls {
...
+		const _target = toVector3( target );
+
+		this._targetEnd.copy( _target );
+		this._sphericalEnd.setFromVector3( _v3.subVectors( _position, _target ) );
```
yomotsu:
```suggestion
		this._sphericalEnd.setFromVector3( _position.sub( _target ) );
```

src/camera-controls.js
FMS-Cat: yeah first I thought about it, but I'd been afraid of argument count explosion.

FMS-Cat:
Note: **The PR causes API compatibility break** in following methods:
- `setLookAt( positionX, positionY, positionZ, targetX, targetY, targetZ, enableTransition )`
- `lerpLookAt( positionAX, positionAY, positionAZ, targetAX, targetAY, targetAZ, positionBX, positionBY, positionBZ, targetBX, targetBY, targetBZ, x, enableTransition )`
- `setPotiion( positionX, positionY, positionZ, enableTransition )`
- `setTarget( targetX, targetY, targetZ, enableTransition )`

src/camera-controls.js
yomotsu: How about the following:

separate this into simpler methods. for the position, and for the target.
We could use the existing `moveTo` methods for the position and `lookat` to the target from the position.

thus, something like this
```
	setLookAt( targetX, targetY, targetZ, enableTransition ) {

		const _position = this.getPosition( _v3a );
		const _target = _v3b.set( targetX, targetY, targetZ );

		this._targetEnd.copy( _target );
		this._sphericalEnd.setFromVector3( _position.sub( _target ) );
		this._sanitizeSphericals();

		if ( ! enableTransition ) {

			this._target.copy( this._targetEnd );
			this._spherical.copy( this._sphericalEnd );

		}

		this._needsUpdate = true;

	}
```

then, users can do this
```
control.moveTo( 5, 6, 7, true );
control.setLookAt( 2, 2, 2, true );
```

yomotsu: er, sorry forget about this, `moveTo` is not for the position...
but I prefer simpler functions like that

yomotsu:
let's think about simpler separating function later.
looks better now and marge it for now

FMS-Cat:
Thanks! :pray:
Should I mark the change as a patch release?

yomotsu:
@FMS-Cat 
right, please up the vertion +0.0.1
maybe:
```
$ npm run build
$ npm run release
$ git commit -m 'build'
```
then
```
$ npm version patch
$ git push
```

FMS-Cat:
Done 🙌

yomotsu:
Thanks! just publish on npm as well

Issue #18, opened 2018-11-30
Event emission feature


FMS-Cat:
## Summary

I want `cameraControls.on( ... )`.
More exactly, I need `cameraControls.on( 'navigate', ( event ) => { ... } )` for now.
Do you have any implementation idea?

## Context

I'm planning to implement auto-pilot using this `camera-controls`.
My requirements are:

- Do auto-pilot initially
- When interaction happens on it, stops the auto-pilot

I know I can implement "move from here to there" on itself using some animation stuff,
but I think it's too overpowered for it.
So, instead, I want `cameraControls.on( 'navigate', ( event ) => { ... } )`-like thing, to achieve halt the auto-pilot when user attempts to move camera.

FMS-Cat:
I think using EventEmitter would be good?
https://www.npmjs.com/package/eventemitter3

FMS-Cat:
Currently I'm imagining implementing these events:
- `navigatestart` : when `mousedown` or `touchstart` occured
- `navigate` : when `mousemove` or `touchmove` occured
- `navigateend` : when `mouseup` or `touchend` occured
- `update` : when camera is updated

FMS-Cat:
I'm wondering the word `navigate` would be proper in this situation

FMS-Cat:
I decided to clone `THREE.EventDispatcher` instead. I'll do pr soon

FMS-Cat:
oh forgot to close this

PR #19, opened 2018-11-30
Implement event dispatcher

FMS-Cat:
See #18.

`CameraControls` now inherits [`EventDispatcher`](https://github.com/FMS-Cat/camera-controls/blob/537d6f5a3703efe8f7d128ea7e0156fa79b79f64/src/event-dispatcher.js)
(The source is completely same as https://github.com/mrdoob/three.js/blob/master/src/core/EventDispatcher.js)

We initially have these events:

- `controlstart`
  when `mousedown` or `touchstart` is occured

- `control`
  when `mousemove` or `touchmove` is occured

- `controlend`
  when `mouseup` or `touchend` is occured

- `update`
  When the camera is updated via controls or transition

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -61,6 +61,15 @@ const cameraControls = new CameraControls( camera, renderer.domElement );
 - `.truckSpeed`: Default is `2.0`. speed of drag for truck and pedestal.
 - `.verticalDragToForward`: Default is `false`. same as `.screenSpacePanning` in three.js's OrbitControls.
 
+## Events
+
+Using `addEventListener( eventname, function )` you can subscribe to these events.
+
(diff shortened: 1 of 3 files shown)


yomotsu:
Thanks for your PR again. I definitely agreed using mrdoob's EventDispatcher.
Seems not bad. I approved to merge this.

FMS-Cat:
Thanks? I also want to see your opinion about these events, is it excessive to implement / is there any other events that would be cool to be there?

yomotsu:
I think should start with the minimum requirements. If someone needs, then we would consider and add. Too many features may cause confusion.

FMS-Cat:
Okay!
I'm gonna update the README and merge it

PR #22, opened 2018-12-17
dollyToCursor

FMS-Cat:
### Public property: `dollyToCursor`

When it's true, dolly using mouse wheel / touch makes it zoom toward its cursor position / touch position.

out


yomotsu:
thanks! looks good!
Merge this for now and I will clean up unnecessary console.log and some👍

PR #24, opened 2018-12-20
dollyToCursor: now it can think about minDistance / maxDistance

FMS-Cat:
Before

before mp4

After

after mp4


yomotsu:
Thanks! marged

PR #25, opened 2018-12-26
add type definitions

FMS-Cat:
Added type definitions under `/types` .
It also adds `@types/three` to its dependency (the first dependency of the repository).


yomotsu:
Thanks! Very helpful!

PR #26, opened 2018-12-26
better type definition

FMS-Cat:
oooops
I just noticed that I should refer this type definition:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/43f46f62429ad12ce1312b9d48d8bdd0528ba4ef/types/three/three-core.d.ts#L1093-L1126


FMS-Cat:
it isn't critical change so we have not to publish this change immediately though

package.json
```diff
@@ -7,7 +7,7 @@
   "repository": "yomotsu/camera-controls",
   "jsnext:main": "dist/camera-controls.module.js",
   "module": "dist/camera-controls.module.js",
-  "types": "src/types/camera-controls.d.ts",
+  "types": "types/camera-controls.d.ts",
```
FMS-Cat: 😐

FMS-Cat:
done.
I'm too noob at TypeScript, forgive me

yomotsu:
Merged. Thanks!

Issue #20, opened 2018-12-05
README: The word `rotX` and `rotY` might be ambiguous


FMS-Cat:
Usually we use the word "rotate X" for rotation around X axis, whereas the document uses rotation around Y axis.
I think we should use `theta` and `phi` instead for here?

https://github.com/yomotsu/camera-controls/blob/73e8d4b9b1fd5db736ba7bfebc3649eb4f26826e/readme.md#L75-L82

yomotsu:
I think rotationX and Y are easier to understand for beginners. maybe we can change to

```
 #### `rotate( theta, phi, enableTransition )` 
  
Rotate azimuthal angle(theta) and polar angle(phi). `theta` and `phi` are in radian. `enableTransition` is in a boolean.

If your camera is Y-up, `theta` is rotationX and `phi` is rotationY.
  ```

FMS-Cat:
I just revisited here!
It's now gonna be no longer documentation only issue, might affect to its API.

I'm planning to add two properties that can change response of rotation via controls.
Should I use `phiSpeed` / `thetaSpeed` instead of `rotateXSpeed` / `rotateYSpeed` ?

yomotsu:
We have used `phiSpeed` and `theta` in the code and added terms for angles to the readme as well.
https://github.com/yomotsu/camera-controls#terms

This issue can be closed now👍
thanks!

PR #27, opened 2018-12-27
Add property that can change rotate speed

FMS-Cat:
### New properties

`phiSpeed` : Change speed of rotation control in phi direction
`thetaSpeed` : Change speed of rotation control in theta direction

Here's a discussion about these property names ("phi" and "theta") : https://github.com/yomotsu/camera-controls/issues/20


yomotsu:
Looks good! Thanks!

PR #28, opened 2018-12-27
DollyToCursor:  It's now not raycaster/object dependant

FMS-Cat:
Previously, translation of `_target` occurs in `dollyToCursor` procedure via mouse wheel events could be happened more than once within a frame, especially under low frame rate conditions.
This PR will fix this problem, since `_target` translation operation is now in `update()` !

### DIFF
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -6,7 +6,6 @@ let _v3A;
 let _v3B;
 let _xColumn;
 let _yColumn;
-let _raycaster;
 let _sphericalA;
 let _sphericalB;
 const EPSILON = 0.001;
@@ -30,7 +29,6 @@ export default class CameraControls extends EventDispatcher {
 		_v3B = new THREE.Vector3();
 		_xColumn = new THREE.Vector3();
 		_yColumn = new THREE.Vector3();
-		_raycaster = new THREE.Raycaster();
 		_sphericalA = new THREE.Spherical();
 		_sphericalB = new THREE.Spherical();
 
@@ -79,6 +77,8 @@ export default class CameraControls extends EventDispatcher {
 		this._position0 = this.object.position.clone();
 		this._zoom0 = this.object.zoom;
 
+		this._dollyControlAmount = 0;
+		this._dollyControlCoord = new THREE.Vector2();
 		this._needsUpdate = true;
 		this.update();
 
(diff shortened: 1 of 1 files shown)


src/camera-controls.js
```diff
@@ -791,6 +785,28 @@ export default class CameraControls extends EventDispatcher {
 
 		}
 
+		if ( this._dollyControlAmount !== 0 ) {
+
+			if ( this.object.isPerspectiveCamera ) {
+
+				const direction = _v3A.copy( _v3A.setFromSpherical( this._sphericalEnd ) ).normalize().negate();
```
yomotsu:
```suggestion
 				const direction = _v3A.setFromSpherical( this._sphericalEnd ).normalize().negate();
```

src/camera-controls.js
```diff
@@ -791,6 +785,28 @@ export default class CameraControls extends EventDispatcher {
...
+				const planeX = new THREE.Vector3().copy( direction ).cross( _v3B.set( 0.0, 1.0, 0.0 ) ).normalize();
```
yomotsu: maybe `_v3C` would be preferred rather than `new THREE.Vector3()`

src/camera-controls.js
```diff
@@ -791,6 +785,28 @@ export default class CameraControls extends EventDispatcher {
...
+				const planeY = _v3B.crossVectors( planeX, direction );
+				const worldToScreen = this._sphericalEnd.radius * Math.tan( this.object.fov / 360.0 * Math.PI );
```
yomotsu: maybe
```suggestion
 				const worldToScreen = this._sphericalEnd.radius * Math.tan( this.object.fov * THREE.DEG2RAD * 0.5 );
```

FMS-Cat: `THREE.Math.DEG2RAD` . what a cool!

src/camera-controls.js
```diff
@@ -791,6 +785,28 @@ export default class CameraControls extends EventDispatcher {
...
+				const prevRadius = this._sphericalEnd.radius - this._dollyControlAmount;
+				const lerpRatio = ( prevRadius - this._sphericalEnd.radius ) / this._sphericalEnd.radius;
+				const cursor = this._targetEnd.clone()
```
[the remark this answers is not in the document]
FMS-Cat: ✔️

FMS-Cat:
I think now I solved all of your reviews. thank you for review 👍

FMS-Cat:
wait

FMS-Cat:
ok I forgot to initialize `_v3C` thanks, now it's confirmed to work

yomotsu:
Merged. Thanks!

PR #29, opened 2019-01-02
modify types path

1016482011:
types folder is not under the src, perhaps you mean `types/camera-controls.d.ts`?


yomotsu:
Merged! Thanks for your contribution.

PR #30, opened 2019-01-03
modify export and update return type in camera-controls.d.ts

1016482011:
Hi, friend, I find some small problems when I apply the awsome cameraControl into my ts project. 
 1. _camera-controls.d.ts_ use the different export with the _camera-controls.js_, which cause I can not import the `CameraControls` in typescript project. 

2.  the public method _update_ should return `boolean` instead of  `void`

### DIFF
--- a/types/camera-controls.d.ts
+++ b/types/camera-controls.d.ts
@@ -1,7 +1,7 @@
 import * as THREE from 'three';
 import { EventDispatcher } from './event-dispatcher';
 
-export class CameraControls extends EventDispatcher {
+export default class CameraControls extends EventDispatcher {
   // static methods
   static install(libs: {THREE: THREE});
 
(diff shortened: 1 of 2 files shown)


yomotsu:
Thanks again! 
Actually, this has been already resolved at another PR: 
https://github.com/yomotsu/camera-controls/pull/26/files#diff-c74ac60224ef1d1a89832447342274ed and it should work. Looks good?

1016482011:
Sure,It  works well, thanks

FMS-Cat:
@1016482011 I'm so sorry about my bad, hope newer d.ts work fine :bow:

1016482011:
@FMS-Cat You are a nice man,  the  .d.ts help me much in ts project

PR #31, opened 2019-01-07
@types/three should be in devDependencies

FMS-Cat:


JvanderHeide:
Was just looking in to this. +1

yomotsu:
Merged. Thanks!

PR #32, opened 2019-01-15
strict=true says: Member 'THREE' implicitly has an 'any' type

FMS-Cat:


yomotsu:
merged! thanks

FMS-Cat:
thanks!

PR #36, opened 2019-03-15
Add boundary feature

### LINKED ISSUE #34: Truck boundaries limit
Hello! Thank you for putting the effort to have this. It is great.

I've been trying to modify it a bit in order to implement bound limits to the **truck** function, but I couldn't find a "nice" way to put it together. I always mess up with something.
 
I am working in a map that needs to be panned within the boundaries oif the model,  but nothing I do does the purpose.

Any tip or advice?

FMS-Cat:
An attempt to resolve #34

- New property: `.boundary`
  - Is a `THREE.Box3` encloses the camera target into a certain region
- New property: `.boundaryFriction`
  - Friction ratio of the boundary, can tweak boundaries behavior when the camera target is in collision
- New property: `.boundaryEnclosesCamera`
  - Whether camera position also should be enclosed in the boundary or not
- plus
  - Update `examples/basic.html`
  - Update README

### DIFF
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -914,30 +914,30 @@ export default class CameraControls extends EventDispatcher {
 
 		 // See: https://twitter.com/FMS_Cat/status/1106508958640988161
 
-		const t1 = _v3B.copy( offset ).add( position ); // target
-		const tc = this.boundary.clampPoint( t1, _v3C ); // clamped target
-		const dtc = tc.sub( t1 ); // t1 -> tc
-		const dtcl = dtc.length(); // length of dtc
+		const newTarget = _v3B.copy( offset ).add( position ); // target
+		const clampedTarget = this.boundary.clampPoint( newTarget, _v3C ); // clamped target
+		const deltaClampedTarget = clampedTarget.sub( newTarget ); // newTarget -> clampedTarget
+		const deltaClampedTargetLength = deltaClampedTarget.length(); // length of deltaClampedTarget
 
-		if ( dtcl === 0.0 ) { // when the position doesn't have to be clamped
+		if ( deltaClampedTargetLength === 0.0 ) { // when the position doesn't have to be clamped
 
 			return position.add( offset );
 
-		} else if ( dtcl === offsetLength ) { // when the position is completely stuck
+		} else if ( deltaClampedTargetLength === offsetLength ) { // when the position is completely stuck
 
 			return position;
 
 		} else if ( friction === 0.0 ) {
 
-			return position.add( offset ).add( dtc );
+			return position.add( offset ).add( deltaClampedTarget );
 
 		} else {
 
-			const offsetFactor = 1.0 + friction * dtcl * dtcl / offset.dot( dtc );
+			const offsetFactor = 1.0 + friction * deltaClampedTargetLength * deltaClampedTargetLength / offset.dot( deltaClampedTarget );
 
 			return position
 				.add( _v3B.copy( offset ).multiplyScalar( offsetFactor ) )
-				.add( dtc.multiplyScalar( 1.0 - friction ) );
+				.add( deltaClampedTarget.multiplyScalar( 1.0 - friction ) );
 
 		}
 
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -709,7 +709,10 @@ export default class CameraControls extends EventDispatcher {
 
 	setBoundary( box3 ) {
 
-		this._boundary.copy( box3 );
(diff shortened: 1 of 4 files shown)


src/camera-controls.js
```diff
@@ -882,6 +902,45 @@ export default class CameraControls extends EventDispatcher {
 
 	}
 
+	_encloseToBoundary( position, offset, friction ) {
```
yomotsu: Do we need the third argument? we could use this. `boundaryFriction` in the method.
```suggestion
	_encloseToBoundary( position, offset ) {
```

src/camera-controls.js
```diff
@@ -882,6 +902,45 @@ export default class CameraControls extends EventDispatcher {
...
+		if ( offsetLength === 0.0 ) { // sanity check
+
+			return position;
+
+		}
+
+		const t1 = _v3B.copy( offset ).add( position ); // target
+		const tc = this.boundary.clampPoint( t1, _v3C ); // clamped target
+		const dtc = tc.sub( t1 ); // t1 -> tc
+		const dtcl = dtc.length(); // length of dtc
```
yomotsu: Full-form name is preferable rather than abbreviated.

yomotsu:
I sent an additonal PR https://github.com/yomotsu/camera-controls/pull/37
Could you take a look that please?

src/camera-controls.js
```diff
@@ -804,6 +813,17 @@ export default class CameraControls extends EventDispatcher {
 		this.object.position.setFromSpherical( this._spherical ).add( this._target );
 		this.object.lookAt( this._target );
 
+		if ( this.boundaryEnclosesCamera ) {
+
+			this._encloseToBoundary(
```
yomotsu: I think just `this.boundary.clampPoint` should be enough at here
```diff
-			this._encloseToBoundary(
-				this.object.position.copy( this._target ),
-				_v3A.setFromSpherical( this._spherical ),
-				1.0
-			);
+			this.boundary.clampPoint(
+				this.object.position,
+				this.object.position
+			);
```

src/camera-controls.js
FMS-Cat: I used mathematic oriented notation here, and also scope is pretty short so I thought it's fine,
but if I need to rename these, I'd choose `newTarget` , `clampedTarget` , `deltaClampedTarget` , `deltaClampedTargetLength`

src/camera-controls.js
FMS-Cat: I prefer the current one. I don't like how the camera moves on the edge when its "friction" is zero

...also I kinda want to change the term "friction" ? How we should call him

src/camera-controls.js
yomotsu: That makes sense. But if you could rename to those, should be better for other users who may read the code🙏

src/camera-controls.js
yomotsu: Sure keep it as is then👍

src/camera-controls.js
FMS-Cat: Okay 'n' done

FMS-Cat:
Types!!! I forgot that, thanks!

src/camera-controls.js
yomotsu: Thanks a lot!

yomotsu:
I guess we should switch to TS entirely sometime soon.

src/camera-controls.js
```diff
@@ -694,6 +707,17 @@ export default class CameraControls extends EventDispatcher {
 
 	}
 
+	setBoundary( box3 ) {
+
+		this._boundary.copy( box3 || new THREE.Box3(
```
FMS-Cat: How about this one?
Added default property for the `setBoundary` method

yomotsu: Looks good!

src/camera-controls.js
```diff
@@ -882,6 +914,47 @@ export default class CameraControls extends EventDispatcher {
...
+			return position;
+
+		}
+
+		 // See: https://twitter.com/FMS_Cat/status/1106508958640988161
+
+		const newTarget = _v3B.copy( offset ).add( position ); // target
+		const clampedTarget = this.boundary.clampPoint( newTarget, _v3C ); // clamped target
+		const deltaClampedTarget = clampedTarget.sub( newTarget ); // newTarget -> clampedTarget
+		const deltaClampedTargetLength2 = deltaClampedTarget.lengthSq(); // squared length of deltaClampedTarget
```
yomotsu: Nice found! this should much faster than `length()`!

yomotsu:
merged! many thanks.

PR #37, opened 2019-03-16
boundary must have the setter

yomotsu:
Users may set `boundaries` at out of 'target'. then `target` will eventually locates out of bounding-box.
For example, user may set the `boundaries` as `( -100, -100, -100 ), ( -90, -90, -90 )`

This PR is to reduce the probrem.

### DIFF
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -58,7 +58,7 @@ export default class CameraControls extends EventDispatcher {
 		this.maxAzimuthAngle = Infinity; // radians
 
 		// Target cannot move outside of this box
-		this.boundary = new THREE.Box3(
+		this._boundary = new THREE.Box3(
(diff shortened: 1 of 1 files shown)


src/camera-controls.js
```diff
@@ -703,6 +703,14 @@ export default class CameraControls extends EventDispatcher {
 
 	}
 
+	setBoundary( box3 ) {
```
yomotsu: This can be `set boundary( box3 ) {}`.

Which one would be better...?

src/camera-controls.js
```diff
@@ -703,6 +703,14 @@ export default class CameraControls extends EventDispatcher {
...
+
+		this._boundary.copy( box3 );
+		this._boundary.clampPoint( this._targetEnd, this._targetEnd );
```
FMS-Cat: ah cool

PR #39, opened 2019-04-11
Add viewport support

FMS-Cat:
Add viewport support for the CameraControls.

When you set viewport to your renderer, the `dollyToCursor` option does not work properly,
since it uses canvas' `getBoundingClientRect` .
The PR adds a new `.setViewport()` method to `CameraControls` .
The viewport is not required but if you set this
you can make the `dollyToCursor` works properly even if renderer's viewport is set.

### WIP

- [x] README
- [x] Type definitions

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -276,6 +276,31 @@ Return its current position.
 
 ---
 
+#### `setViewport( vector4? )`
+
+Set (or unset) the current viewport.
+
+Set this when you want to use renderer viewport and [`.dollyToCursor`](#properties) feature at the same time.
+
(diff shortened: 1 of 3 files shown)


FMS-Cat:
This one would work properly:

https://gist.github.com/FMS-Cat/9808f07722a9c04f2efc7af2cc0af1cf

yomotsu:
Thanks. looks good so far.
Would you mind adding [the example](https://gist.github.com/FMS-Cat/9808f07722a9c04f2efc7af2cc0af1cf) later? Simplified one would be better.

FMS-Cat:
sure! > an example

PR #40, opened 2019-04-19
New example: viewport.html

FMS-Cat:
Add an example that can be used to examine `cameraControls.setViewport()` .

image


yomotsu:
merged. many thanks!

PR #41, opened 2019-04-19
Fix touch pan/truck issue

FMS-Cat:
When you attempt to pan or truck the camera using touch screen,
it causes unintended behavior under certain conditions: when the document is scrolled.
This PR fixes the issue.

### DIFF
--- a/src/camera-controls.js
+++ b/src/camera-controls.js
@@ -291,15 +291,15 @@ export default class CameraControls extends EventDispatcher {
 				if ( scope._state === STATE.TOUCH_DOLLY_TRUCK ) {
 
 					// 2 finger pinch
-					const dx = _v2.x - event.touches[ 1 ].pageX;
-					const dy = _v2.y - event.touches[ 1 ].pageY;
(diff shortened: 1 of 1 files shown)


src/camera-controls.js
```diff
@@ -291,15 +291,15 @@ export default class CameraControls extends EventDispatcher {
 				if ( scope._state === STATE.TOUCH_DOLLY_TRUCK ) {
 
 					// 2 finger pinch
-					const dx = _v2.x - event.touches[ 1 ].pageX;
```
FMS-Cat: Is there any reason we should use `pageX` / `pageY` instead of `clientX` / `clientY` at here?

FMS-Cat: All changes belongs to the PR are just replacing these `pageX` / `pageY` into `clientX` / `clientY`

yomotsu: Good finding. you are right.
I have not tested with laid out canvas elements...
thanks!

PR #42, opened 2019-04-23
Fix build script

FMS-Cat:
`npm run build` didn't work on Windows so I fixed it


yomotsu:
Very sorry about that. thanks!

PR #52, opened 2019-08-01
Implement functionality for any camera.up vector

Jesusbill:
This PR adds the functionality for any camera.up vector. It follows the implementation of `OrbitControls` from the official repository of threejs.
There is also a minor modification suggested for the case when dollyToCursor is enabled and the view direction is exactly equal to the camera.up vector, for which dollyToCursor does not function.

All the examples in the repository have been tested with `camera.up = (0, 0, 1)` which is the main alternative to the default settings and everything works as expected.

Not being a Javascript experienced user it is possible that the implemented code could be improved, although pretty straightforward. I leave that to the discretion of the reviewers.


yomotsu:
Thanks LTGM👍
Just merged

Jesusbill:
Great!

PR #55, opened 2019-09-03
fix setLookAt from any camera.up

Jesusbill:
Implementing the module in my project where I use `camera.up = (0, 0, 1)` I realized that the positioning of the camera in the `setLookAt` method was not modified accordingly to all the `sphericalEnd.setFromVector3` calls in the PR that implemented the utility of the camera controls for any camera.up vector, resulting, thus, in an erroneous position placement of the camera.

This behavior can be reproduced by eliminating line 57 from the `camera-up.html` example
```
camera.position.set( 0, 5, 0 );
```
and by adding the following at line 62 just after the definition of cameraControls
```
cameraControls.setLookAt( 0, 5, 0, 0, 0, 0, true );
```

This PR fixes this bug and updated the example to test the correct behavior.

### DIFF
--- a/examples/camera-up.html
+++ b/examples/camera-up.html
@@ -54,12 +54,12 @@ const scene  = new THREE.Scene();
 const camera = new THREE.PerspectiveCamera( 60, width / height, 0.01, 100 );
 
 camera.up.set( 0, 0, 1 );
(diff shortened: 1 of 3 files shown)


yomotsu:
Nice finding.
Thanks!

PR #67, opened 2020-01-10
add trackpad config

yomotsu:
see https://github.com/yomotsu/camera-controls/pull/66#issuecomment-572880522

### DIFF
--- a/examples/config.html
+++ b/examples/config.html
@@ -45,6 +45,15 @@
 		<option value="NONE">CameraControls.ACTION.NONE</option>
 	</select>
 	<br>
+	cameraControls.trackpad.two =
+	<select onchange="cameraControls.trackpad.two = CameraControls.ACTION[ this.value ]">
+		<option value="ROTATE">CameraControls.ACTION.ROTATE</option>
+		<option value="TRUCK">CameraControls.ACTION.TRUCK</option>
+		<option value="DOLLY" selected>CameraControls.ACTION.DOLLY</option>
+		<option value="ZOOM">CameraControls.ACTION.ZOOM</option>
+		<option value="NONE">CameraControls.ACTION.NONE</option>
+	</select>
+	<br>
 
 	cameraControls.touches.one =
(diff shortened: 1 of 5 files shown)


yomotsu:
@andrewplummer Could I make trackpad feature as experimental? I mean, when you use the feature, you need to enable the trackpad config.

```
cameraControls.enableUnstableTrackpadConfig = true;
cameraControls.trackpad.two = ACTION.TRUCK;
```

Because in order to detect trackpad move, it relies on `wheelDeltaY`, which is [a deprecated value](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent) of MouseEvent.

andrewplummer:
Yes absolutely! Feels like an experimental feature to me.

yomotsu:
Thanks for understanding!
It's gonna be available as part of the lib at the next release.

andrewplummer:
Update: I'm finally able to confirm this using my mouse, and it seems my assumption (taken from this SO post here: https://stackoverflow.com/questions/10744645/detect-touchpad-vs-mouse-in-javascript) is incorrect... at least for my Logitech Master 2s `wheelDeltaY` is always 3x `deltaY` so it is detecting as a trackpad. I've looked at everything else I can think of on the wheel event, and it seems that nothing works, unfortunately.

I think this PR still adds value as it allows trucking via the wheel event (this could be useful for a mouse wheel that has X axis as well). For now I think I'm going to have to simply use a UI toggle for zoom vs truck to handle this instead.

yomotsu:
@andrewplummer Thanks for your report. So could we just remove the trackpad feature? Since the code is kinda hacky, it might cause other problems as well 😢

andrewplummer:
What if we remove the `isTrackpad` code and config, and instead simply do `mouseButtons.wheel = ACTION.TRUCK`?

yomotsu:
@andrewplummer okay, I'm working on it at https://github.com/yomotsu/camera-controls/pull/77

andrewplummer:
Awesome, thank you!

Issue #70, opened 2020-01-16
An accessor cannot be declared in an ambient context.


flashpuller:
Error in Angular 8 project using Angular CLI. (Note it still supporting typescript<3.6.0):

ERROR in node_modules/camera-controls/dist/CameraControls.d.ts:6:16 - error TS1086: An accessor cannot be declared in an ambient context.

6     static get ACTION(): Readonly<typeof ACTION>;
                 ~~~~~~
node_modules/camera-controls/dist/CameraControls.d.ts:50:9 - error TS1086: An accessor cannot be declared in an ambient context.

50     set phiSpeed(speed: number);
           ~~~~~~~~
node_modules/camera-controls/dist/CameraControls.d.ts:51:9 - error TS1086: An accessor cannot be declared in an ambient context.

51     set thetaSpeed(speed: number);
           ~~~~~~~~~~
node_modules/camera-controls/dist/CameraControls.d.ts:52:9 - error TS1086: An accessor cannot be declared in an ambient context.

52     get boundaryEnclosesCamera(): boolean;
           ~~~~~~~~~~~~~~~~~~~~~~
node_modules/camera-controls/dist/CameraControls.d.ts:53:9 - error TS1086: An accessor cannot be declared in an ambient context.

53     set boundaryEnclosesCamera(boundaryEnclosesCamera: boolean);
           ~~~~~~~~~~~~~~~~~~~~~~

yomotsu:
Not sure but maybe because of the version of TS.
The d.ts is [created with TS3.7](https://github.com/yomotsu/camera-controls/blob/bec69da9b4787af50c587dc0efb9885499e34ff7/package.json#L20).

Could you try with 3.7 in your project?

omararturo:
Same problem!

yomotsu:
Sorry for the inconvenience.
I'll downgrade to TS 3.0.x. then, the error should be removed, since getter/setter in d.ts has been added at TS 3.7.0.

yomotsu:
The problem should be solved with camera-controls v1.18.2.

PR #72, opened 2020-01-17
downgrade to TS3.0.x

yomotsu:
see https://github.com/yomotsu/camera-controls/issues/70


yomotsu:
Also I tried [downlevel-dts](https://github.com/sandersn/downlevel-dts) with TS 3.7, which is released recently but seems still has some unexpected behavior.

Maybe next time.

PR #82, opened 2020-02-27
three should be in peerDependencies

FMS-Cat:
`three` should be in peerDependencies I think because:
- camera-controls has no use without three.js
- its type definitions depends on three

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,9 @@
     "three": "^0.112.1",
     "typescript": "3.2.1"
   },
+  "peerDependencies": {
+    "three": "^0.112.1"
+  },
   "scripts": {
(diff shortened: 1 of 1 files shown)


yomotsu:
Make sense!

package.json
```diff
@@ -19,6 +19,9 @@
     "three": "^0.112.1",
```
yomotsu: Iis this no longer needed then?
```suggestion
```

FMS-Cat: peerDependencies are not included in the repo as dependencies.

yomotsu: okay thanks!

yomotsu:
merged!

PR #83, opened 2020-02-27
update ts version and downlavel d.ts version

yomotsu:
- update ts version to 3.8.2
- downlevel d.ts version to 3.4. see https://github.com/yomotsu/camera-controls/issues/70
- use type-only option


yomotsu:
Thanks for reviewing❤️

PR #91, opened 2020-04-03
Update onMouseWheel listener

jeanremy:
Hi,

I was trying to remove the listener on the wheel, because sometimes the canvas is just a part of the page. While it works well when ACTION.NONE is specified, it just blocks the scroll when you are on top of the canvas (because of the e.preventDefault). So before entering the function, we can just check if controls are enabled and if the user config is not NONE.
I'm not very familiar with TypeScript, I hope this change is acceptable.

Thank for your job by the way !


yomotsu:
Thanks for your contribution!
That makes sense!

PR #92, opened 2020-04-05
Add zoom state to toJSON() and fromJSON()

### LINKED ISSUE #87: Include zoom in JSON state?
Hi, I was thinking the zoom probably should be included in the JSON state, since it's critical for the orthographic camera.

For completeness, I could also imagine adding an accessor to the camera controls to get this value. Though that isn't essential. I'm able to pick the zoom from the camera directly.

What do you think?

paulmelnikow:
This is critical when using an orthographic camera.

Closes #87

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -835,7 +835,7 @@ export class CameraControls extends EventDispatcher {
 		bb.min.y -= paddingBottom;
 		bb.max.x += paddingRight;
 		bb.max.y += paddingTop;
-		
+
 		const bbSize = bb.getSize( _v3B );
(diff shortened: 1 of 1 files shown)


src/CameraControls.ts
```diff
@@ -835,7 +835,7 @@ export class CameraControls extends EventDispatcher {
 		bb.min.y -= paddingBottom;
 		bb.max.x += paddingRight;
 		bb.max.y += paddingTop;
-		
+
```
paulmelnikow, line 838: The linter complained about this. Is the linting running in CI?

yomotsu: uh oh thanks, I should set up the CI...

yomotsu:
Thanks! merged!

Issue #100, opened 2020-06-09
I expect have a controls can support zoomed and rotated with the camera position


liuyehua:
orbitcontrols rotated center awalys in center of screen。。。if it can zoomed and rotated with the current camera position in screen。。。

yomotsu:
Hi liuyehua,
Sorry, I don't get your point. Could you explain that in more detail, please?

PerryRylance:
@yomotsu I think he might be asking about zooming in on the mouse pointer, which is what I'm also here to ask about.

Is there any way to get the camera to zoom (and translate) about the mouse pointer on mousewheel, rather than about the center of the screen?

A bit like how Google Maps' zoom works.

Great camera, by the way.

yomotsu:
I see, thanks!
I'm working on at https://github.com/yomotsu/camera-controls/pull/103

`dollyToCursor` feature is already available tho

Jesusbill:
@yomotsu what is the difference between dolly and zoom? I already use `dollyToCursor` to zoom to mouse cursor.

yomotsu:
See here. I personally prefer using dolly too!
https://github.com/yomotsu/camera-controls#dolly-vs-zoom

PerryRylance:
@Jesusbill dolly to cursor isn't as useful for orthographic cameras

PerryRylance:
So, I've got the following code, which I've run the numbers on and it will work, however for some reason the camera moves in a very unpredictable way when I add this to `zoomInternal`

```
var zoomTowards = {
	x: _this._camera.left + (x / window.innerWidth)  * (_this._camera.right - _this._camera.left),
	y: _this._camera.top  + (y / window.innerHeight) * (_this._camera.bottom - _this._camera.top)
};

var center = {
	x: _this._camera.left + (_this._camera.right - _this._camera.left) / 2,
	y: _this._camera.top  + (_this._camera.bottom - _this._camera.top) / 2
};

var position = new THREE.Vector3(
	center.x + ((zoomTowards.x - center.x) * zoomScale), 
	center.y + ((zoomTowards.y - center.y) * zoomScale), 
	0
);

this.moveTo(position.x, position.y, this._camera.position.z);
```

Am I doing something wrong here in terms of interacting with your code @yomotsu ?

Thank you again for your time in this :)

yomotsu:
Ill take a look

yomotsu:
Closing this issue.
Zoom-to-cursor for orthographic has been implemented:
[...]

PR #103, opened 2020-07-01
[WIP] zoom to cursor

yomotsu:
see https://github.com/yomotsu/camera-controls/issues/100

support zoom-to-cursor

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -131,6 +131,7 @@ export class CameraControls extends EventDispatcher {
 	protected _position0: _THREE.Vector3;
 	protected _zoom0: number;
 
+	protected _zoomControlAmount = 0;
 	protected _dollyControlAmount = 0;
 	protected _dollyControlCoord: _THREE.Vector2;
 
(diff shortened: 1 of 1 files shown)


FMS-Cat:
👀

PerryRylance:
Hi @yomotsu , thank you so much for this,

I've rebuilt this at my end and found this works great for perspective cameras, how would we get this to work for Orthographic cameras?

Is that something you could potentially accommodate in this pull request?

yomotsu:
@PerryRylance Actually I'm struggling with this PR,
I guess even the zoom to cursor for perspective cameras does not work correctory and will reject the PR at this moment😢

Hopefully, someone could help..🙏

ebeit303:
Hi @yomotsu , Good Job.!!! 
Mouse pointer zoom works perfect, 

1. I check your examples, while setting infinityDolly as true, you checking with min distance which is always 0, so zoom cannot be done infinitely.
2. While Mouse pointer zoom, the target need to changed as per mouse pointer(camera direction), so the rotation will be done with the newly target not with (0,0,0). It is must if you looking for doing mouse pointer zoom

Other thing are Good..!!!! 
Have a good Day.

yomotsu:
sorry but let me drop this PR for now.
maybe try this again sometime soon.

Issue #115, opened 2020-07-22
How to enable damping for dollying?


flolu:
I couldn't find any options to enable "dolly-damping". (smooth transitions when scrolling with mouse-wheel)
Does `camera-controls` not support it or was I just unable to find it?

yomotsu:
maybe `.dollySpeed`?

flolu:
@yomotsu `.dollySpeed` only adjusts the amount of dollying. However, my goal is to achieve a smooth dolly transition.

See [this example](https://modelviewer.dev/). When scrolling in or out, there is a dolly animation. 
But [here](https://yomotsu.github.io/camera-controls/examples/basic.html) the dollying is instantaneos, without an animation.

Can I achieve this "smooth transition" with `camera-controls` or this out of the scope of this project?
Thanks for your help :smiley:

yomotsu:
ah, do you mean mouse wheel in Windows?
so far, mouse wheel dolly does not support "damping", as you can see the second argument is falsy here:
https://github.com/yomotsu/camera-controls/blob/dev/src/CameraControls.ts#L273
(The second argument represents whether to use damping or not)

Because it may conflict native smooth scroll on mac trackpad wheel.
(if the user uses macOS, wheel dolly works smoothly)

if there is a nice way to detect whether the native wheel is smooth or not, I could set the second argument depends on that.

yomotsu:
Thanks for the code!
Seems like, native inertial scrolling can be disabled by user setting on macOS, and it may also problem
(I should research and test more tho, but it takes a little bit time)

for a quick workaround, you can set
```
cameraControls.mouseButtons.wheel = CameraControls.ACTION.ZOOM
```
to disable wheel action and make your own wheel action instead using
https://github.com/yomotsu/camera-controls/blob/dev/src/CameraControls.ts#L376
and set true to the second argument of `dollyTo()` in your code.
(hope it works for you)

If it does not work, let me know.

flolu:
Awesome! I'll look into that tomorrow and report back.
@yomotsu I really appreciate your fast help!

yomotsu:
er, I guess I can set the dumping option internally here
https://github.com/yomotsu/camera-controls/blob/dev/src/CameraControls.ts#L273

I tested that on my mac and not bad actually...

yomotsu:
Closing due to lack of feedback.

VegarRingdalAibel:
@yomotsu 
[...]

PR #116, opened 2020-07-22
enable wheel dolly damping

yomotsu:
see https://github.com/yomotsu/camera-controls/issues/115

Damping factor for dolly had been disabled because of mac native inertial scrolling.
But actually with damping on mac is not bad unless damping is not too low.


yomotsu:
none linear dolly will break dolly-to-cursor
I need to consider more...

mikeskydev:
This works really well for me! Would love to see it as part of the main package.

PR #122, opened 2020-09-01
Change compile target to es6 and update dev dependencies.

ThreeDfish:


yomotsu:
Thanks but, please don't put multiple subjects in one PR.

- Don't change the version per PR. (imagine if the version is increased by every PR)
- Separate dependencies update and build setting.

Although, I would like to keep targeting es5 until the end of this year.

PR #123, opened 2020-09-03
THREE.Math is now THREE.MathUtils

msemtd:
See https://github.com/mrdoob/three.js/pull/18270 and https://github.com/mrdoob/three.js/releases/tag/r113


yomotsu:
I didn't know `THREE.Math` has been moved to `legacy` already.
Thanks!

Issue #124, opened 2020-09-03
'ReferenceError: navigator is not defined' for Next.js apps


nikhilxb:
I'm trying to use this library within my Next.js app, and I'm getting a `ReferenceError: navigator is not defined` thrown by this line, since `navigator` is not defined on the server side.

https://github.com/yomotsu/camera-controls/blob/71ffdc4e797d328bdb70134b3d5832158eddb160/src/CameraControls.ts#L26

nikhilxb:
I found a hack that fixes it.

Broken:
```js
import CameraControls from 'camera-controls';
CameraControls.install({ THREE });
```
Working:
```js
const CameraControls = typeof window !== `undefined` ? require("camera-controls").default : null;
if(CameraControls) CameraControls.install({ THREE });
```

yomotsu:
Ah, that's right.
We should add another condition for SSR detection in camera-controls.

I'll work on it this weekend, or you can make a PR for it if you like.

yomotsu:
I think this should be solved with https://github.com/yomotsu/camera-controls/pull/125

nikhilxb:
Awesome, thanks!

PR #125, opened 2020-09-04
Detect whether node or browser

yomotsu:
refs https://github.com/yomotsu/camera-controls/issues/124

Add a condition to detect running on whether browser or node.


yomotsu:
@nikhilxb does this look good to you?

PR #126, opened 2020-09-07
Updated readme.md to prompt the user to run CameraControls.install()

ACMCMC:
I ran into problems the first time I used camera-controls, because I didn't copy the example usage code, which contains the line `CameraControls.install( { THREE: THREE } )`. This is neccesary for camera-controls to work, but is not explicitly stated. I've added a warning to remind the user about the importance of running that line.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -58,6 +58,17 @@ const cameraControls = new CameraControls( camera, renderer.domElement );
 } )();
 ```
 
+### Important!
+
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks! This should help other users too.

Maybe adding `console.error( 'you must...' )` if `THREE` is undefined in the library would be nice as well?
Anyway I'll merge your PR for now 😄

ACMCMC:
Yes! Adding that would be great, I will take care of that ASAP :)

PR #127, opened 2020-09-08
Added an error message if the user hasn't run install()

ACMCMC:
#126


src/CameraControls.ts
```diff
@@ -153,6 +153,11 @@ export class CameraControls extends EventDispatcher {
 
 		super();
 
+		if (typeof THREE === 'undefined') { //Chech if the user has installed THREE
+			console.error("Three.js is undefined. You must first run CameraControls.install({ THREE: THREE }). Check the docs for further information.")
+			return; //Stop the execution as there will be errors
```
yomotsu, line 158: Could I omit `return` here?
Without initializing properties (such as `this._camera` and others ) causes TS error 😢.
```suggestion
```

ACMCMC: For sure!

yomotsu:
Just merged.
Thanks for your contribution again!

PR #137, opened 2020-09-25
Update get/toJSON/loadJSON methods with focalOffset

nikhilxb:
- Mirror other properties (target, position, etc.) by adding `getFocalOffset()`.
   - Useful because sometimes `setFocalOffset()` values need to be calculated as a function of the current values, which are changed during trucking.
- Preserve focalOffset, focalOffset0 in `toJSON()`/`loadJSON()`.
   - Needed for completeness for recreating camera state.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -427,7 +427,17 @@ Return its current position.
 
 #### `getTarget( out )`
 
-Return its current gazing target which is the center position of the orbit.
+Return its current gazing target, which is the center position of the orbit.
+
+| Name  | Type            | Description |
+| ----- | --------------- | ----------- |
+| `out` | `THREE.Vector3` | The receiving vector |
+
+---
+
+#### `getFocalOffset( out )`
+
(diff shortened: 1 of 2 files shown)


nikhilxb:
Note that this could be a breaking change, because `loadJSON()` currently does not seem to use default values in case `obj.focalOffset`/`obj.position`/etc do not exist. So any it is not possible to use `loadJSON()` with JSON saved from earlier versions.

But this problem is not specific to `focalOffset`, as any adding any new field in the future will result in the same problem. Not sure if it matters to you at this stage, but thought I'd just bring it up.

yomotsu:
Thanks for your first contribution!
Looks perfect!

If you could, would you mind adding `getFocalOffset` to the doc(readme.md) as well?

> Note that this could be a breaking change
I will manage that later👍

nikhilxb:
Done!

yomotsu:
Thank you so much!

yomotsu:
just merged

PR #143, opened 2020-10-09
feat: make `fitToBox()` abort itself when it is given an empty box

FMS-Cat:
It seems doing `fitToBox` against empty boxes completely messes its internal state up since `v1.20.x`.
First of all I don't think people wants to do `fitToBox` against empty boxes.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -974,6 +974,13 @@ export class CameraControls extends EventDispatcher {
 			? _box3A.copy( box3OrObject as _THREE.Box3 )
 			: _box3A.setFromObject( box3OrObject as _THREE.Object3D );
 
+		if ( aabb.isEmpty() )  {
+
+			console.warn( 'camera-controls: fitTo() cannot be used with an empty box. Aborting' );
+			return;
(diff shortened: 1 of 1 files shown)


src/CameraControls.ts
```diff
@@ -974,6 +974,13 @@ export class CameraControls extends EventDispatcher {
 			? _box3A.copy( box3OrObject as _THREE.Box3 )
 			: _box3A.setFromObject( box3OrObject as _THREE.Object3D );
 
+		if ( aabb.isEmpty() )  {
+
+			console.warn( 'camera-controls: fitTo() cannot be used with an empty box. Aborting' );
+			return;
```
FMS-Cat, line 980: I chose warn + early abort. Does it look good for you?

yomotsu:
Thanks for your investigation.

> First of all I don't think people want to do fitToBox against empty boxes.

That's a good point. Now that I think of it, it could happen with default `box3`.

yomotsu (approved):
Approved

yomotsu:
merged! thanks

PR #152, opened 2020-12-07
Bump three r123

### LINKED ISSUE #151: `Quaternion.inverse()` is replaced to `Quaternion.invert()` since r123
### Description

`Quaternion.inverse()` is removed and replaced by `Quaternion.invert()` since r123 (See: https://github.com/mrdoob/three.js/pull/20613)
We uses `Quaternion.inverse()` and we want to remove the use of `inverse()` or Three.js will emit a deprecation warning for each calls of `inverse` in r123.

### Thing I want to make a consensus before I work on this

@yomotsu 
Here's how I resolved this while preserving compatibility with previous versions: https://github.com/pixiv/three-vrm/pull/556/files

```ts
export function quatInvertCompat<T extends THREE.Quaternion>(target: T): T {
  if ((target as any).invert) {
    target.invert();
  } else {
    (target as any).inverse();
  }

  return target;
}
```

Does this look good for you?
If so, I can work on this along with a commit that bumps Three.js to r123

FMS-Cat:
Will resolve #151 

- Three.js is now r123
- `Quaternion.inverse` -> `Quaternion.invert`
    - Compat is implemented for backward compatibility

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1399,9 +1399,9 @@
       "dev": true
     },
     "three": {
-      "version": "0.120.1",
-      "resolved": "https://registry.npmjs.org/three/-/three-0.120.1.tgz",
-      "integrity": "sha512-ktaCRFUR7JUZcKec+cBRz+oBex5pOVaJhrtxvFF2T7on53o9UkEux+/Nh1g/4zeb4t/pbxIFcADbn/ACu3LC1g==",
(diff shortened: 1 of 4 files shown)


yomotsu (approved):
Approved!

yomotsu:
Sorry for the delay and thanks for your PR.
Looks nice as usual. merged.

PR #155, opened 2020-12-21
Removing the need to install.

evandertoorn:


yomotsu:
Thanks for your contribution.
However, I think `install()` is needed for pure ESM.

e.g.
```html
<script type="module">
import * as THREE from 'https://unpkg.com/three@0.123.0/build/three.module.js';
import CameraControls from 'https://unpkg.com/camera-controls@1.25.3/dist/camera-controls.module.js';

CameraControls.install( { THREE } );
<script>
```

yomotsu:
Closing due to lack of feedback.

mikeskydev:
Import maps are coming which would allow this? I've used this package in the past to load both three and camera-controls into my viewer instead of bundling them.

Package: https://www.npmjs.com/package/es-module-shims
Example usage: https://github.com/icosa-gallery/icosa-viewer/blob/8cfc1572071d8f7bccf55c58bc6122dec17019a6/examples/gltf-viewer.html

If not now, I think it would be good to consider when import maps become a common thing.

PR #159, opened 2021-02-17
add fitToSphere support for orthographic camera

ervinoro:
Really a very simple change to use zoom instead of dolly in fitToSphere in the case of orthographic camera.

Additionally, it includes a bit of refactoring:
* use type guards for camera type
* run eslint autofix
* run npm update (outdated eslint had a bug with type guards)

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -1079,7 +1079,6 @@ export class CameraControls extends EventDispatcher {
 		const boundingSphere = isSphere ?
 			_sphere.copy( sphereOrMesh as _THREE.Sphere ) :
 			createBoundingSphere( sphereOrMesh as _THREE.Object3D, _sphere );
-		const distanceToFit = this.getDistanceToFitSphere( boundingSphere.radius );
 
 		this.moveTo(
 			boundingSphere.center.x,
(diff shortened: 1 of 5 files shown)


src/CameraControls.ts
```diff
@@ -222,17 +224,17 @@ export class CameraControls extends EventDispatcher {
 			middle: ACTION.DOLLY,
 			right: ACTION.TRUCK,
 			wheel:
-				( this._camera as THREE.PerspectiveCamera  ).isPerspectiveCamera  ? ACTION.DOLLY :
-				( this._camera as THREE.OrthographicCamera ).isOrthographicCamera ? ACTION.ZOOM :
+				isPerspectiveCamera( this._camera )  ? ACTION.DOLLY :
+				isOrthographicCamera( this._camera ) ? ACTION.ZOOM :
```
yomotsu, line 228: Function call could be slightly costy. Could we just use the property?

ervinoro: I believe that using the type guards instead of type casting improves code clarity and helps to avoid errors. I'm not quite sure about the function call overhead. https://jsbench.me/lyklgbaglm/1 this extremely simplistic benchmark reports the property being actually 17% slower than function call on Firefox, and both as being equal on Chrome (for me). There is some conflicting information online, apparently function calls are slower on Internet Explorer.

I have rebased the PR such that the commit "add type guards for camera type" is last. Feel free to exclude it if you want to keep using the property.

yomotsu: Thanks for understanding, and the test😊
three.js tends to avoid using functions such as https://github.com/mrdoob/three.js/pull/19997#issuecomment-673685741
therefore, let me do the same at this moment.

yomotsu:
Merged.
Thank you for your first contribution!

Houssk:
thanks @ervinoro  !

PR #165, opened 2021-03-29
Upgrade three to 0.126.1

flolu:
Fixes #160

Reason for changes in `src/CameraControls.ts` line `1772-1780`: https://discourse.threejs.org/t/solved-geometry-vertices-is-undefined/3133/2


yomotsu:
Thanks for your contribution! Very helpful😊

Actually, three.js was only for the types in camera-controls. However, as you know, three.js has recently [removed the type from its repo](https://github.com/mrdoob/three.js/pull/21174).

Could we just refer [@types/three](https://www.npmjs.com/package/@types/three) for types?

flolu:
It's a little bit confusing. They have removed all types and Typescript users should now use [`@types/three`](https://www.npmjs.com/package/@types/three).
But there is also [`three-ts-types`](https://github.com/three-types/three-ts-types), which does not seem to have an npm release.

So I think our best option for now is [`@types/three`](https://www.npmjs.com/package/@types/three).

It's honestly kind of sad. I thought they would convert Three.js to a Typescript project at some point. But I don't think there is any hope for that. Would be too much work for the maintainers

yomotsu:
> does not seem to have an npm release.

Don't worry, `three-ts-types` released in the `@types/three`, according to the README in `three-ts-types`.
Would you mind changing the type reference to `@types/three`, if you have some time?

flolu:
I don't think we need to change the type references.

If I try it, I get this error:
```
Cannot import type declaration files. Consider importing 'three' instead of '@types/three'.ts(6137)
```

I think the `@types/<something>` modules are designed in such a way that they automatically link the types to `<something>`

yomotsu:
Ah, I understand! Thanks for the details🙇‍♂️

yomotsu:
Merged! many thanks!

PR #178, opened 2021-05-10
Listen for events of the domElement

farfromrefug:
Not sure why for some events you were listening on the document. My issue is that in the environment i am using i dont have touch events on the document.
Also it seems better to listen on the domElement if you have multiple canvas.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -355,12 +355,12 @@ export class CameraControls extends EventDispatcher {
 
 				this._state = ACTION.NONE;
 
-				document.removeEventListener( 'mousemove', dragging );
+				_this._domElement.removeEventListener( 'mousemove', dragging );
 				// see https://github.com/microsoft/TypeScript/issues/32912#issuecomment-522142969
 				// eslint-disable-next-line no-undef
-				document.removeEventListener( 'touchmove', dragging, { passive: false } as AddEventListenerOptions );
-				document.removeEventListener( 'mouseup',  endDragging );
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your contribution.

`document.addEventListener` can detect mouse events even that is happened outside the browser.
For instance, a user starts dragging from the element, then the user may move the cursor to outside of the browser.

farfromrefug:
@yomotsu can we have an option to listen on one or the other? In my env (outside of html web page, kind of like nodejs) three cameras work but not yours because of that.

yomotsu:
Do you mean you would like to use it in server-side rendering?

this lib already has a sort of [browser detection](https://github.com/yomotsu/camera-controls/blob/1be0ba33bf1935aa222ff38bc820740ab4e46ce0/src/CameraControls.ts#L29).

We could add conditions for your case.
Could I hear the details of your usecase a little bit more?

yomotsu:
Thanks for your explanation. I understand.

How about something like this?
```
( document || this._domElement ).removeEventListener( 'mousemove', dragging );
```

farfromrefug:
@yomotsu sadly this would not work because `document` exists. Simply for now it wont "emit" events for the touches.
But i am starting to think i could "forward" all events to that "document" element.
I will run some tests and report here

farfromrefug:
@yomotsu so i managed to "forward" all my events through the document object and so your lib works without any change!
sorry for the trouble and thank for your time/help

PR #179, opened 2021-05-14
fix: prevent warning on mobile devices

farfromrefug:
On mobile device i was seeing a lot of `Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted`.
This fixes it

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -553,8 +553,10 @@ export class CameraControls extends EventDispatcher {
 			const dragging = ( event: Event ): void => {
 
 				if ( ! this._enabled ) return;
-
-				event.preventDefault();
+				
+				if (event.cancelable) {
+					event.preventDefault();
+				}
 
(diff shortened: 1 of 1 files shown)


src/CameraControls.ts
```diff
@@ -553,8 +553,10 @@ export class CameraControls extends EventDispatcher {
 			const dragging = ( event: Event ): void => {
 
 				if ( ! this._enabled ) return;
-
-				event.preventDefault();
+				
+				if (event.cancelable) {
+					event.preventDefault();
+				}
```
yomotsu, line 559: Would you mind adapting the lint as well?
```suggestion
				
				if ( event.cancelable ) event.preventDefault();
```

yomotsu:
Thanks for another PR. Looks nice!

yomotsu:
Merged.
Many thanks!

PR #183, opened 2021-05-27
feature: seprate domElement and other HTMLElement touch event

FateLocker:


yomotsu:
Sorry for the long delay...
Could I hear about the usecase?

FateLocker:
I am trying to add an analog joystick to the window, which can manipulate the position of the camera; but I hope that the camera-control will not recognize the touch event on the joystick view

yomotsu:
That makes sense. I understand.
But this change could be a breaking change.

How about adding a new property called `ignoreElements[]` or so?

FateLocker:
It could not be better

yomotsu:
Thank you for understanding.
If you are still interested in the PR, I will happy to receive the change.
I can also work on it based on your PR but it may take some time.

yomotsu:
@FateLocker 

Hey, I've added PointerEvent API support.
I think it's easy to ignore events that started from `ignoreElements`.

https://github.com/yomotsu/camera-controls/blob/ac39858d675fd30b831968c5642c9723a7debf8c/src/CameraControls.ts#L264-L278

If the event started from `ignoreElements`, the only we need is just `return` immediately!

FateLocker:
It has only responded to `domElement`, but its normal function is affected by 'isPointerEventsNotSupported'

https://github.com/yomotsu/camera-controls/blob/ac39858d675fd30b831968c5642c9723a7debf8c/src/CameraControls.ts#L790-L791

yomotsu:
That's right. So, we can just ignore the event at the beginning of the handlers.
`onPointerDown`
`onMouseDown`
`onTouchStart`

```js
const onMouseDown = ( event ) => {

  // if event.target is a descendant of ignoreElement, return here

}
```

And do the same for other handlers.

FateLocker:
I mean that after using PointerEvents, other HTMLElement will no longer affect CameraControls.

yomotsu:
Ah, I see!
So you mean, this PR is no longer needed?

yomotsu:
@FateLocker
Thank you for your PR anyway!
I'm looking forward to another PR from you and joining as a contributor!

Issue #187, opened 2021-06-04
[FEATURE REQUEST] option to "handle" zoom in dragging


farfromrefug:
Right now if you dragged while zoomed the dragging is "multiplied" by the zoom factor making it hard to move around.
Could we have an option to "slow down" the dragging while zoomed?
I test it locally and replacing:
```typescript
this.rotate( theta, phi, true );
```
with 

```typescript
this.rotate( theta / _this._zoom, phi / _this._zoom, true );
```
[here](https://github.com/yomotsu/camera-controls/blob/4c4bf6f66c79f03affd7099065d61aa1ec194afa/src/CameraControls.ts#L304 ) gets the right behavior

I can create a PR i just need to figure out how to name that option

yomotsu:
How about make `xxxInternal()` extendable?

```js
class CameraControls {
	constructor() {
		...
	},

	protected _rotateInternal( deltaX: number, deltaY: number ): void => {
		...
	}
}
```
then users can modify internal functions:

```js
class MyCameraControls extends CameraControls {
	constructor() {
		super();
	}

	_rotateInternal( deltaX: number, deltaY: number ): void => {
		// what ever
	}
}
```

farfromrefug:
@yomotsu i was actually about to ask you to move all functions as class functions so that they can be extendable :D

yomotsu:
We also need to move [`elementRect`](https://github.com/yomotsu/camera-controls/blob/4c4bf6f66c79f03affd7099065d61aa1ec194afa/src/CameraControls.ts#L302-L303) (and maybe others) to be private too, but It's okay for me.

Do you have some time to make a PR for it?
I could do that this weekend tho.

farfromrefug:
@yomotsu let me know what you think https://github.com/yomotsu/camera-controls/pull/188

yomotsu:
Handler methods can be customized by https://github.com/yomotsu/camera-controls/pull/189
Closing the issue.

PR #188, opened 2021-06-04
chore: refactor to allow method override

farfromrefug:
as discussed here https://github.com/yomotsu/camera-controls/issues/187

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -116,8 +116,6 @@ export class CameraControls extends EventDispatcher {
 	mouseButtons: MouseButtons;
 	touches: Touches;
 
-	cancel: () => void = () => {};
-
 	protected _enabled = true;
 	protected _camera: _THREE.PerspectiveCamera | _THREE.OrthographicCamera;
 	protected _yAxisUpSpace: _THREE.Quaternion;
@@ -158,6 +156,11 @@ export class CameraControls extends EventDispatcher {
 
 	protected _needsUpdate = true;
 	protected _updatedLastTime = false;
+
+	protected _dragStartPosition: _THREE.Vector2;
+	protected _lastDragPosition: _THREE.Vector2;
+	protected _dollyStart: _THREE.Vector2;
+	protected _elementRect: _THREE.Vector4;
 
 	constructor(
 		camera: _THREE.PerspectiveCamera | _THREE.OrthographicCamera,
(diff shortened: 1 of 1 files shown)


src/CameraControls.ts
```diff
@@ -159,6 +157,11 @@ export class CameraControls extends EventDispatcher {
 	protected _needsUpdate = true;
 	protected _updatedLastTime = false;
 
+	protected _dragStartPosition: _THREE.Vector2;
+	protected _lastDragPosition: _THREE.Vector2;
+	protected _dollyStart: _THREE.Vector2;
```
yomotsu, line 162: these 3 do not need to be private. let's make it as it is

src/CameraControls.ts
```diff
@@ -116,8 +116,6 @@ export class CameraControls extends EventDispatcher {
 	mouseButtons: MouseButtons;
 	touches: Touches;
 
-	cancel: () => void = () => {};
-
```
yomotsu, line 120: Could we focus only `xxxInternal()` in this PR?

src/CameraControls.ts
farfromrefug: i follow the same rule as the other ones by making them protected. Couldn't it be useful for a extended method to be able to access thoses?

src/CameraControls.ts
farfromrefug: Well i thought the best was to do them all once and for all :p

src/CameraControls.ts
yomotsu: these 3 are only used in DOM event handlers.
Maybe we could expose them later.

farfromrefug: @yomotsu ok will change that

src/CameraControls.ts
```diff
@@ -159,6 +157,11 @@ export class CameraControls extends EventDispatcher {
 	protected _needsUpdate = true;
 	protected _updatedLastTime = false;
 
+	private _dragStartPosition: _THREE.Vector2;
+	private _lastDragPosition: _THREE.Vector2;
+	private _dollyStart: _THREE.Vector2;
+	private _elementRect: _THREE.Vector4;
```
yomotsu, line 163: er, sorry I meant,
`protected` for `_elementRect`
and revert `_dragStartPosition`, `_lastDragPosition` and `_dollyStart` as scoped

farfromrefug: @yomotsu actually we cant have `dragStartPosition`, `_lastDragPosition` and `_dollyStart` as scoped. They are used my multiple class functions. We can have then as real private with `#` if you want

src/CameraControls.ts
```diff
@@ -239,595 +242,604 @@ export class CameraControls extends EventDispatcher {
...
-					const truckX    = ( this.truckSpeed * deltaX * targetDistance / elementRect.w );
-					const pedestalY = ( this.truckSpeed * deltaY * targetDistance / elementRect.w );
-					if ( this.verticalDragToForward ) {
+		// if ( this._domElement ) {
+		this._dragStartPosition = new THREE.Vector2() as _THREE.Vector2;
+		this._lastDragPosition = new THREE.Vector2() as _THREE.Vector2;
+		this._dollyStart = new THREE.Vector2() as _THREE.Vector2;
+		this._elementRect = new THREE.Vector4() as _THREE.Vector4;
+		// }
+
```
yomotsu, line 251: Then this part going to be...
```suggestion
		this._elementRect = new THREE.Vector4();

		if ( this._domElement ) {

			const dragStartPosition = new THREE.Vector2() as _THREE.Vector2;
			const lastDragPosition = new THREE.Vector2() as _THREE.Vector2;
			const dollyStart = new THREE.Vector2() as _THREE.Vector2;
```

and scoped `cancelDragging`, `onMouseDown`, and other handers.

farfromrefug: @yomotsu if i may i would really like those to be extendable too. I actually already have the need for it :P Especially `onMouseDown`. May i ask why you want them scoped?

yomotsu: I understand. However would you mind separating this PR to make the diff smaller?
- xxxInternal for this PR
- DOM event handlers for another

farfromrefug: ok i can wont have time to do it now though.

yomotsu:
@farfromrefug BTW, Could I hear your usecase of custom DOM event handlers?

farfromrefug:
@yomotsu i need to listen for touch start and touch end but i need to make sure it is done before camera controls does it. I could add a touchstart event listener but then i am not sure it is(or i am i sure it is not)  done before camera controls.

yomotsu:
I think you could do it already.
Doesn't it work for your case?

```
class MyCameraControls extends CameraControls () {
  constructor( ... ) {
    xxx.addEventListener( 'touchstart', yourHandler );
    
    super( ... );
  }
}
```

farfromrefug:
@yomotsu i would like not to add a second event listener.  Also i tried `controlstart` but it was not working in all cases (if i remember correctly.
I will close this PR as the other PR is kind of enough for now. I got almost all i need with it.
I ll let you decide what to do with the DOM part.
Thanks a lot for your work!

PR #194, opened 2021-06-28
feat: Add a Getter and Setter for the camera

Fennec-hub:
Adding a Getter and Setter for the camera giving the user the ability to switch cameras during run time without having to create a new `CameraControls` object for each new camera.


src/CameraControls.ts
```diff
@@ -616,6 +587,51 @@ export class CameraControls extends EventDispatcher {
...
+	set camera(camera: _THREE.PerspectiveCamera | _THREE.OrthographicCamera) {
+		this._camera = camera;
+
+		// camera up
+		this._yAxisUpSpace = new THREE.Quaternion().setFromUnitVectors( this._camera.up, _AXIS_Y );
+		this._yAxisUpSpaceInverse = quatInvertCompat( this._yAxisUpSpace.clone() );
+
+		// rotation
+		this._spherical = new THREE.Spherical().setFromVector3( _v3A.copy( this._camera.position ).applyQuaternion( this._yAxisUpSpace ) );
+		this._sphericalEnd = this._spherical.clone();
```
yomotsu, line 603: I think `this._yAxisUpSpace`, `this._spherical`, and other settings should be shared with cameras, weather than making new instances.

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "noUnusedParameters": true,
     "strictNullChecks": true,
     "strictFunctionTypes": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "strictPropertyInitialization": false,
```
yomotsu, line 19: I prefer using `?` at class member fields if really needed.

yomotsu:
Thanks for your contribution. Could I hear your use case for this?

tsconfig.json
Fennec-hub: Yeah! I was hesitant on this one, typescript can be a real pain sometimes.

Fennec-hub:
Hi,

First of all thanks for for sharing this amazing work, saved me a lot of troubles, in regard of my use case, I'm trying to build a `Three.js` editor and I need to switch between `Orthographic` and `Perspective` cameras, beside I think it's very common to change the orbit controls camera, it can be useful in many scenarios.

Since we are talking editors, please take a look at this [Orientation Gizmo](https://github.com/Fennec-hub/ThreeOrbitControlsGizmo), I will definitely adapt it with your `camera-controls` to use it with my editor, may be we can merge it with your project.

One last thing, the static installation method you're using is troubling me, is there a reason why you don't just use Rollup `externals`, please take a look at my second commit and, If you you don't mind, tell me what you think.

Thanks again.

yomotsu:
Thanks for the reply and use. I understand.
Switching the camera type must be useful in that case,  like Blender.

The feature should be added however would you mind changing some I commented on above?

---

`CameraControls.install( { THREE } )` is necessary for pure JS module using CDN.

e.g.
```
import * as THREE from 'https://cdn.com/three';
import CameraControls from 'https://another-cdn.com/camera-controls';
```

JS cannot resolve dependencies unless using import-maps

tsconfig.json
yomotsu: So would you mind reverting tsconfig and using `!`?

src/CameraControls.ts
```diff
@@ -1,4 +1,4 @@
-import type * as _THREE from 'three';
+import * as THREE from 'three';
```
yomotsu, line 1: Please use three only for type reference😢

Fennec-hub: Well understood,
I will reorder everything accordingly, just give me some times and sorry for the disturbance, this is actually my first PR.

src/CameraControls.ts
yomotsu: @Fennec-hub
I think `set camera` should only change `this._camera` and the CameraControls instance should keep the same state, such as `this._yAxisUpSpace`, `this._spherical`, and other props.

otherwise, `set camera` could change props accidentally, I guess...

yomotsu (changes requested):

yomotsu:
Thanks for understanding!

Fennec-hub:
You get a point, it make no sense creating new instances of `this._spherical = new THREE.Spherical()` and `this._yAxisUpSpace = new THREE.Quaternion()` for each camera change.

I'll remove this fork, cleaning up the mess I just made, create a new one and start a new the PR, hopping it will meet your standards.

Thanks for your patients and again sorry for any disturbance  :sweat_smile:

PR #195, opened 2021-06-28
feat: Add a Getter and Setter for the camera

Fennec-hub:
Give the user the ability to change the camera during runtime, removing the need of creating a new `CameraControls` instance for each new camera, especially useful for switching between `PerspectiveCamera` and `OrthographicCamera`.


src/CameraControls.ts
```diff
@@ -616,6 +587,51 @@ export class CameraControls extends EventDispatcher {
...
+		// Touch controls config
+		this.touches = {
+			one: ACTION.TOUCH_ROTATE,
+			two:
+				isPerspectiveCamera( this._camera )  ? ACTION.TOUCH_DOLLY_TRUCK :
+				isOrthographicCamera( this._camera ) ? ACTION.TOUCH_ZOOM_TRUCK :
+				ACTION.NONE,
+			three: ACTION.TOUCH_TRUCK,
+		};
+	}
```
yomotsu, line 633: Thanks for organizing the PR.
However, I realized that `set camera` can be just a setter of the `_camera` without any side effects.

```suggestion
	set camera( camera: _THREE.PerspectiveCamera | _THREE.OrthographicCamera ) {
		this._camera = camera;
		this._needsUpdate = true;
	}
```

And other setting changes can be done in the user space.

e.g.
```
cameraControls.camera = anotherCamera;
// touches.two have to be changed, but I don't want to change for touches.one;
cameraControls.touches.two = whatever;
```

Fennec-hub: For the `touches` and `mouseButtons`, I agree with you, they're already exposed to the user.

> I realized that set camera can be just a setter of the _camera without any side effects.

Are you sure about that, I mean what if the new camera has a different `position`, `up` or `zoom`, especially with `OrthographicCamera` where the zoom is important.

If you want to remove those settings anyway, it's fine by me, I'll commit you're updates.

Thanks

yomotsu: If users need to change `position`, `zoom`, and other props, I think the user can make another `CameraControls` instance rather than reusing it.

It would be nice if the user could freely set whether the position and zoom should move immediately or smoothly.
However, if the `set camera` changes the position or zoom, it cannot be handled.

yomotsu:
@Fennec-hub 
I'm very sorry, but I've made another PR. https://github.com/yomotsu/camera-controls/pull/198
Because this PR still contains unnecessary changes...
But very thankful!

Fennec-hub:
Don't worry, I perfectly understand and I should be the one thanking you for the great work.

yomotsu:
Thanks for understanding. I will close the PR at this time (and will merge  #198 instead).
But, I'm looking forward to having another PR from you.

PR #196, opened 2021-06-28
Add shift, alt & ctrl for mouse button actions

ketourneau:


yomotsu:
Thanks for your contribution.

Before merging this, let me confirm some points.
Do you need all of them? If not, could we start with small changes as possible?
Otherwise, We also need to think about what if a user pressing `ctrl` and `alt` at the same time or so...

Also, would you mind changing the doc (readme), please?
https://github.com/yomotsu/camera-controls#user-input-config

ketourneau:
Hi @yomotsu,

We only need shiftLeft action so I update the PR.
I also updated readme & example.

yomotsu:
Thanks! merged!

Issue #185, opened 2021-05-28
Feature request: rotateTo only on azimuth or polar


yannbriancon:
Use case: 

- I need to rotate to a specific azimuthAngle with transition without changing the polarAngle target.

Problem:

- I cannot use rotate because I don't know the delta between the current end azimuthAngle and the new one I want (_sphericalEnd is protected)
- I cannot use rotateTo because I don't know the current end polarAngle (_sphericalEnd is protected)

Proposed solutions:

- Expose _sphericalEnd to let the developer compute his delta
- (Preferred) Allow rotate and rotateTo to take only azimuthAngle or polarAngle
  - If azimuthAngle or polarAngle is undefined, keep the same _sphericalEnd

yomotsu:
I understand. That would be good to have.

We already have a setter called `azimuthAngle` and `polarAngle`.
https://github.com/yomotsu/camera-controls/blob/4c4bf6f66c79f03affd7099065d61aa1ec194afa/src/CameraControls.ts#L754

Currently, it doesn't have the second argument thus it immoderately set the angle. maybe the setter could have the second argument as `enableTransition: boolean`.

Another idea is adding new methods like:
```
rotateAzimuthTo( azimuthAngleTo, enableTransition )
```
But I'm not confident about the name of the methods...(any suggestions...?)

mijofr:
My suggestion would simply be to allow rotate or rotateTo to be called with something like controls.rotateTo(MATH.PI/2, null, true); A null indicates no change from current.

yannbriancon:
@yomotsu @mijofr 
Here is a PR with my proposal https://github.com/yomotsu/camera-controls/pull/201.
I am happy to discuss other solutions.
Cheers!

yomotsu:
@yannbriancon's PR has been merged and published into npm. (many thanks!)
Closing the issue.

PR #201, opened 2021-07-16
Feature: rotateTo on one angle only

yannbriancon:
ref: https://github.com/yomotsu/camera-controls/issues/185

Add the possibility to rotate to an azimuth angle without changing the polar angle target (and vice versa)

- Allow use of undefined for rotateTo azimuthAngle and polarAngle parameters to keep the current value in sphericalEnd.theta or  sphericalEnd.phi


yomotsu:
Thank you for your PR.
I personally think... I would like the argument to be required if it is not optional.
Could we have separated methods, which might be called `rotateAzimuthTo( azimuthAngleTo, enableTransition )` or so?

yannbriancon:
Yes definitely!
I will update my PR later ;)

yannbriancon:
@yomotsu I updated the PR with the new methods!

yomotsu:
Thanks! LGTM!

yomotsu:
Just merged.
This PR has been marked as next released and will be published on NPM this weekend.

PR #205, opened 2021-08-18
truckXDisabled and truckYDisabled options

edsilv:
This PR is to enable users to constrain trucking by x or y axis.
Our use case is that we're using camera-controls to navigate various data visualisations, where data is arranged by x and y axis. For example, we have a date view where we only want the user to be able to pan horizontally, not vertically in order to create a sliding timeline.
We also want to surface the ability to enable/disable pan on the x/y axis to the user via a menu.


yomotsu:
Thanks for your contribution.
What is the use-case of this PR?
Do you mind filling in the description section, please?

edsilv:
@yomotsu Sure, have filled out the description.
I did look at the boundary option, but it's not quite what we needed - `truckXDisabled` and `truckYDisabled` gives us specific control over an axis as opposed to having to calculate the bounds of a boundary and give it either a 0 width (`truckXDisabled`) or 0 height (`truckYDisabled`).
I thought to make them disabled setters instead of `truckXEnabled` in that they're falsey by default. Sort of like TrackballControls [noPan](https://threejs.org/docs/#examples/en/controls/TrackballControls.noPan)

yomotsu:
Thanks for the details.

But I'm nervous to add new options.

- What if another user wants `truckXSpeed` and `truckYSpeed` separately...? `truckXDisabled` could be `truckXSpeed = 0`...?
- The lib already has [user-input-config](https://github.com/yomotsu/camera-controls#user-input-config). Should this be a config...?

and so on...

Maybe you could extend the lib to change the truck dragging behavior for now.

```js
class MyCameraControls extends CameraControls {

  constructor( ... ) {
    ...
  },


  _truckInternal( deltaX, deltaY, dragToOffset ) {

    // disable Y truck
    super._truckInternal( deltaX, 0, dragToOffset );

  }

}

```

edsilv:
I tried the extension approach:

https://codesandbox.io/s/camera-controls-extended-uwtgl?file=/index.html

But getting:

`(intermediate value)._truckInternal is not a function`

yomotsu:
If you don't use TS, super may not work, since the Class was transpile to Function due to ES5 support.
https://codesandbox.io/s/camera-controls-extended-forked-yobuu?file=/index.html

edsilv:
Thanks @yomotsu - that's working perfectly.
I think this is a good approach more generally anyway for any other modifications I need to make.

yomotsu:
Good to hear that.
And thank you for the PR.
Unfortunately, it was not merged at this time, but looking forward to having the next one!

Issue #206, opened 2021-08-21
[FEATURE] Make transitions thenable 


looeee:
**Is your feature request related to a problem? Please describe.**

1. I need to disable parts of the interface while the camera is animating, e.g. using `controls.rotateTo`. 

``` js
disableUI();
await controls.rotateAzimuthTo(Math.PI/4, true);
enableUI();
```

2. I would like to chain animations e.g:

``` js
await controls.dollyTo(5, true);
await controls.rotateAzimuthTo(Math.PI/4, true);
await controls.dollyTo(2, true);
// ... etc.
```

**Describe alternatives you've considered**

I've tried doing this using the event listeners `wake` and `sleep` however the problem is:
a) it's hard to distinguish an animation from a user interaction
b) due to damping `sleep` fires far too late

yomotsu:
Hi @looeee,
I'm happy to have your comment, the author of the book, as well as a frequent contributor of three.js!

If you'd like to control a CameraControls instance as tweens, you could combine a tween library,

e.g.
https://yomotsu.github.io/camera-controls/examples/easing.html

---

- CameraControls relies on a damping animation, as you understand, and not time-based animations(like tween).
- Because, damping animation can combine multiple input dynamically, unlike time-based fixed animation.
- Although, CameraControls can be applied to time-based fixed animation as well.
- (But damping animation must be ignored during fixed animations.)
- To apply the fixed animation, use a Tween library such as tween.js, GSAP, or whatever.

What do you think...?

looeee:
> I'm happy to have your comment 

😊 

> If you'd like to control a CameraControls instance as tweens, you could combine a tween library,

Yeah for sure, I'm using Popmotion at the moment. Actually it was GSAP's new version that inspired me to make this request, their tweens return promises now which makes writing animations much cleaner. No more `onComplete` everywhere. 

> Because, damping animation can combine multiple input dynamically, unlike time-based fixed animation.

Yeah, this is the difficult one for knowing when an animation has completed. I explored the code a bit more since I made this issue and I see why it would be a bit hard to implement. 

The problem is `sleep` firing too late because of damping - usually it fire a number of seconds after it "looks" like the camera has stopped. 

[...]

PR #207, opened 2021-08-29
Feat/add rest and transitionstart events

looeee:
This PR adds the new `rest` event and corresponding `restEpsilon` parameter discussed in #206. 

Also added a new `rest-and-sleep` example to show the difference between the two events, and updated the readme. 

While making the example, I discovered there was no way to reliably disable the buttons when starting a new transition using one of the buttons (which call `.rotateTo` and `.dollyTo`) - so I added an additional event, `transitionstart`. This fires whenever any kind of transition is started, which can be either a user input or a method with `enableTransition = true`. 

I originally wanted `transitionstart` to fire only for a method with `enableTransition = true`, but _not_ for user control. However, internally user controls use methods such as `.rotate` so there's currently no way to differentiate these. Probably that's ok though. 

I checked in `dist/` by mistake, let me know if you want me to remove it to make the code easier to review.


yomotsu:
Thanks for the PR! I'll take a look.

I should have indicated that it's better to ignore the dist folder...but it's okay.

src/CameraControls.ts
```diff
@@ -111,6 +112,8 @@ export class CameraControls extends EventDispatcher {
 	verticalDragToForward = false;
 
 	boundaryFriction = 0.0;
+	
+	restEpsilon = 0.0025;
```
yomotsu, line 116: Might be `restThreshold`...?

looeee: OK, done. Also remove dist/

yomotsu:
Thanks! Just merged!

PR #208, opened 2021-08-30
Feat: make transitions thenable

looeee:
Building on #207 this makes transitions thenable, so you can now create complex camera animations using async/await:

``` js
async function complexTransition() {
        disableUI(true);
	await cameraControls.rotateTo( Math.PI / 2, Math.PI / 4, true );
	await cameraControls.dollyTo( 3, true );
	await cameraControls.fitToSphere( mesh, true );
	disableUI(false);
}
```

This will rotate the camera, then dolly, and finally fit to the bounding sphere of the `mesh`. It will also disable UI elements while the animation is playing. 

Create a new example `await-transitions` to show how it works. All methods that take the `enableTransitions` param are included. 

I also adjusted the default `restThreshold` to be 0.01 as this looks better to me after more testing.

### DIFF
--- a/examples/await-transitions.html
+++ b/examples/await-transitions.html
@@ -1,0 +1,187 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width">
+<title>=^.^=</title>
+<link rel="stylesheet" href="./style.css">
+</head>
+<body>
+<div class="info">
+	<p><a href="https://github.com/yomotsu/camera-controls">GitHub repo</a></p>
+	<button type="button" onclick="reset()">Reset</button>
+	<button type="button" onclick="transitionSequenceA()">Transition Sequence A</button>
+	<button type="button" onclick="transitionSequenceB()">Transition Sequence B</button>
+
+	<p>Transition promises will resolve when the camera slows to less than <code>controls.restThreshold</code> units in a single frame. You can set this value here</p> 	<label><input oninput="cameraControls.restThreshold = this.value" value="0.01" type="number" min="0.001" max="1" step="0.001"> Rest Threshold</label>
+
+</div>
+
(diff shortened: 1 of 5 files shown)


yomotsu:
Thanks for another PR.
I'll take a look tomorrow or later (I have covid vaccine side effects 🥵)

looeee:
Oh damn! Still, better than real covid symptoms I guess 😅 Get well soon - and no rush to check this. 

Regarding the PR - there's quite a bit of repeated code when creating the promise in each method:

``` ts
return new Promise( ( resolve ) => {

	const onResolve = () => {

	    this.removeEventListener( "rest", onResolve );

	    resolve();

    };

    this.addEventListener( "rest", onResolve );

} );
```

This could probably be moved into a util or something. 

Also the example is a bit lazy haha - currently I'm just throwing a bunch of random numbers so it sometimes looks good, sometimes not great. 

It could be made more interesting and also I wanted to experiment with changing the `.restThreshhold` and `.dampingFactor` during the transitions, something like:

``` ts
async function complexTransition() {
	const initialDF = cameraControls.dampingFactor;
	
    disableUI(true);
    
    // increase the threshold to make the animations transition faster
    cameraControls.restThreshold = 0.05;
    
  	// Slow rotate
	cameraControls.dampingFactor = initialDF / 2;
	await cameraControls.rotateTo( Math.PI / 2, Math.PI / 4, true );
	
	// Super slow dolly
	cameraControls.dampingFactor = initialDF / 4;
	await cameraControls.dollyTo( 3, true );
	
	// Back to normal speed and threshold for the final fit
	cameraControls.dampingFactor = initialDF;
	cameraControls.restThreshold = 0.01;
		
	await cameraControls.fitToSphere( mesh, true );

	disableUI(false);
}
```

Haven't tested this yet though.

looeee:
Awesome! Glad you're feeling better 💪

PR #213, opened 2021-09-15
Add orbit point method fix #130

### LINKED ISSUE #130: Set orbit point different from camera target point
I'm trying to make it so that when a user clicks a part of my 3D model, the camera's orbit point is changed to that part's centerpoint _without_ moving the camera (which `setTarget()` does). The goal is for the camera to orbit around that part, even if the user pans/trucks the camera.

Is there a way to change the orbit point?

The solution from a couple years ago (https://github.com/yomotsu/camera-controls/issues/8) of setting `controls._target` and `controls._targetEnd` moves the camera and does not work after the user pans/trucks the camera.

ketourneau:
We find the solution to set orbit point without changing the camera position

```ts
setOrbitPoint( targetX: number, targetY: number, targetZ : number ) {
    _xColumn.setFromMatrixColumn( this._camera.matrixWorldInverse, 0 );
    _yColumn.setFromMatrixColumn( this._camera.matrixWorldInverse, 1 );
    _zColumn.setFromMatrixColumn( this._camera.matrixWorldInverse, 2 );

    let cameraToPoint = new THREE.Vector3( targetX, targetY, targetZ ).sub( this._camera.position );
    _xColumn.multiplyScalar( cameraToPoint.x );
    _yColumn.multiplyScalar( cameraToPoint.y );
    _zColumn.multiplyScalar( cameraToPoint.z );

    _v3A.copy( _xColumn ).add( _yColumn ).add( _zColumn );
    _v3A.z = _v3A.z + this.distance;

    this.setFocalOffset( - _v3A.x, _v3A.y, - _v3A.z, false );
    this.moveTo( targetX, targetY, targetZ, false );
}
```

### DIFF
--- a/examples/focal-offset.html
+++ b/examples/focal-offset.html
@@ -19,6 +19,9 @@
 	<button onclick="cameraControls.setFocalOffset( 2, 2, 0, true )">setFocalOffset( 2, 2, 0 )</button>
 	<button onclick="cameraControls.setFocalOffset( 0, 0, 2, true )">setFocalOffset( 0, 0, 2 )</button>
 	<br>
+	<button onclick="cameraControls.setOrbitPoint( 2, 0, 0, true )">setOrbitPoint( 2, 0, 0 )</button>
+	<button onclick="cameraControls.setOrbitPoint( 0, 0, 0, true )">setOrbitPoint( 0, 0, 0 )</button>
+	<br>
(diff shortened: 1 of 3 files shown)


yomotsu:
Fantastic! Thanks for the PR.
just merged!

PR #231, opened 2021-10-28
Set parameter to optional

SubZtep:
Calling `setBoundary()` without parameters causes _`Expected 1 arguments, but got 0.ts(2554)`_ error.

This PR simply makes the parameter optional – as [described in the doc](https://github.com/yomotsu/camera-controls#setboundary-box3-).


yomotsu:
Thanks! that's right!

yomotsu:
Merged! Many thanks!

PR #275, opened 2022-06-21
Add camera speed on movement

klich3:
feature: Can set/change speed for camera move / walk


yomotsu:
Thanks for your PR.
However, I don't get the point.
Could you please elaborate?
- What does 'walk' means?
- What is this PR for, and usecase?

Also please don't include generated code (via `npm run build`). It causes too many changes and makes review difficult.

klich3:
Hi there, for now I use camera-controls module on two projects, like Matterport "click and go" on some 3D space on clicking on some wall or floor camera is walking like on first person. 
This I mean walk. Sorry for include `build` I do some test and push it accidently.

[Video demo](https://www.loom.com/share/b1286bd6572841ff9c08c7006939ce3a) 
on Init. I have normal speed.
on half of the video value is set at 0.01.

klich3:
The GSAP library is so heavy only for animate forward moviment, I think many people will look / use this parameter for same propose.

Now you bind `restThreshold` for rotations and pan / truck, I think isn't correct. Because if you use first person controls and apply that you slows key's responses / moviement and rotation "looking around". 
With separete posibility you can rotate with normal speed and truck / pan camera with other speed.

yomotsu:
I agree that GSAP is too much.
But there are several options for tween libs. You can have a light one such as tween.js, even you can make it yourself, [like me](https://github.com/yomotsu/tween-values)!

We must be careful to add a new feature to `update()`. `update()` will be called every frame. As long as we have another option, I would like to avoid adding things to `update()`.

Does this make sense...?

yomotsu:
Thanks for understanding.
Let me close this PR, however, I really appreciate your contribution.
I look forward to having another one from you.

Issue #276, opened 2022-06-21
fitToBox broken with Z-axis Up


rxmarccall:
## Describe the bug
The fitToBox function does not result in a Box3 being properly focused by the camera when `camera.up` is set to use the Z-axis. The issue is particularly noticeable when using a Box3 that has a much larger vertically than wide. (vertical rectangle)

## To Reproduce

1. Set the ThreeJS camera's `up` property to: `new Vector3(0, 0, 1);`
2. Call `cameraControls.updateCameraUp();`
3. Create a Box3 that's primarily vertical, see code below for values to try.
4. Call `this.cameraControls.fitToBox(tempBox3, false);`
### Code

```js
//init Camera
this.camera = new THREE.PerspectiveCamera(30, this.divRef.current.offsetWidth / this.divRef.current.offsetHeight, 0.1, 1000);

// Set camera's up to use Z axis
this.camera.up.set(0, 0, 1);
this.camera.position.z = 50;

// Create camera controls
this.cameraControls = new CameraControls( this.camera, this.renderer.domElement );

// Update camera controls 'Up' since we made camera's up use the Z axis
this.cameraControls.updateCameraUp();

// Create a vertical test Box3 for demonstrating fit to box issue and run fitToBox after a delay
setTimeout(() => {
const tempBox3 = new THREE.Box3();
tempBox3.setFromPoints([
    new THREE.Vector3(-10, -40, 0),
    new THREE.Vector3(10, 40, 5),
]);

const boxHelper = new THREE.Box3Helper(tempBox3);
this.scene.add(boxHelper);

this.cameraControls.fitToBox(tempBox3, false);
}, 2000);
```

### Live example
NA

## Expected behavior
fitToBox should focus / fit a Box3 correctly regardless of Camera's up axis or a Box3's dimensions.

## Screenshots or Video
Result with camera's up set to Z axis
![image](https://user-images.githubusercontent.com/1920223/174898839-d9f355c1-2af0-497f-a5df-e33a70208c9c.png)
Result with camera's standard up orientation
![image](https://user-images.githubusercontent.com/1920223/174898889-5bb3704f-26cb-4cc0-ad8b-e08b12301c77.png)

## Platform:

 - Device: [Desktop]
 - OS: [MacOS]
 - Browser: [Chrome]

yomotsu:
Thanks for your report. I'll take a look.

yomotsu:
Reproduced the problem. marked as a bug.

[...]

PR #296, opened 2022-08-09
fix fitToBox in non-Yup space

ntrfrnc:
I found that fitToBox not working properly in non-Yup space. There is closed issue #276 about the same problem but in some circumstances it is still broken. This PR should fix two issues I found, and I believe make fitToBox work properly in every scenario in non-Yup space.

### Two problems I found

1. When one dimension of box is smaller than two others - fitting to most of the box sides is broken.
2. When fit from up or bottom - box is fitted always to the same size, despite changing azimuth angle by 90 deg.

### Reproduce first issue
1. open https://yomotsu.github.io/camera-controls/examples/camera-up.html
3. open console
2. run:

```
const box = new THREE.Box3(
  new THREE.Vector3(0, 0, 0),
  new THREE.Vector3(0.1, 0.5, 0.1),
);

scene.add(new THREE.Box3Helper(box));
cameraControls.rotateTo(Math.PI * 0.5, Math.PI * 0.5, true);
cameraControls.fitToBox(box, true);
```

### Reproduce second issue

Same as first issue but instead of two last lines run:

```cameraControls.rotateTo(Math.PI * 0.5, 0, true);cameraControls.fitToBox(box, true);```

And to see azimuth angle that is fitted properly:

```cameraControls.rotateTo(0, 0, true);cameraControls.fitToBox(box, true);```

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -1637,13 +1637,15 @@ export class CameraControls extends EventDispatcher {
 		promises.push( this.rotateTo( theta, phi, enableTransition ) );
 
 		const normal = _v3A.setFromSpherical( this._sphericalEnd ).normalize();
-		const rotation = _quaternionA.setFromUnitVectors( normal, _AXIS_Z ).multiply( this._yAxisUpSpaceInverse );
+		const rotation = _quaternionA.setFromUnitVectors( normal, _AXIS_Z );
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your patch!
I confirmed the problem and it works perfectly in your branch.
I will merge this and publish the patched version.

PR #309, opened 2022-09-27
two finger rotate&zoom (TOUCH_DOLLY_ROTATE, TOUCH_ZOOM_ROTATE)

taejaehan:
I'm using the camera-controls library just fine. Thank you.

I checked and used the option to change rotation and zoom with a two-finger touch. 
(recently merged options : TOUCH_DOLLY_ROTATE, TOUCH_ZOOM_ROTATE)
However, if I fix one finger and rotate it, it seems to work well, but I think it works strangely if I use both hands to move two fingers at the same time. 

So, I added the logic to get a vector of two fingers and rotate them. 
This logic works with TOUCH_DOLLY_ROTATE and TOUCH_ZOOM_ROTATE.

Please review. thank you.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -21,6 +21,7 @@ import {
 	roundToStep,
 	infinityToMaxNumber,
 	maxNumberToInfinity,
+	getRotateAngle
 } from './utils/math-utils';
 import { extractClientCoordFromEvent } from './utils/extractClientCoordFromEvent';
 import { notSupportedInOrthographicCamera } from './utils/notSupportedInOrthographicCamera';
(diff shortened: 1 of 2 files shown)


yomotsu:
Thanks for your PR.
Regarding "two fingers and rotate", Do you mean moving 2 fingers in a circular motion, like the below image?
image

If so, I'm not sure this action should rotate horizontally.

taejaehan:
Yes, "two fingers and rotate" means to rotate like the image you attached.

I made it because it needed to rotate horizontally as well as rotating horizontally by moving one finger left and right.
And it works better with two-finger touch using both hands as shown in the image below.
스크린샷 2022-09-28 오후 12 22 20

You said "two fingers and rotate" should be rotated in the camera-up direction, I'm not sure what that means. Are you talking to the global Y-axis?

yomotsu:
> global Y-axis?

Sort of. I think the behavior for the mosion is dependent on projects. some users might think of camera-up rotation, thus I believe we should keep it as it is.

If you really wish the implication, What about starting with extending `_rotateInternal()` in your project, like https://github.com/yomotsu/camera-controls/pull/205#issuecomment-901984036

taejae-han:
Okay. I understand it will be different for each project.
I'll try it later as you suggested.
Thanks for making a great library

yomotsu:
Thank you for your understanding.
At this time, we don't merge the PR, but look forward to another PR from you.
Thanks again. Let me close this PR.

FYI: Vector2 has some useful methods such as `dot()` `cross()` `length()` and `angle()` to replace your utility code. try using them: https://threejs.org/docs/#api/en/math/Vector2

PR #331, opened 2022-12-22
Fix check for dolly state when dragging

dsafa:
This PR fixes an issue where it would zoom instead of dolly if `this._state` had other flags set.
You can observe the current behavior [in the examples](https://yomotsu.github.io/camera-controls/examples/basic.html).
If you hold down the middle mouse button and also hold down the right mouse button at the same time, then drag, the camera will start to `zoom + truck` instead of `dollying + truck`.

Fixed by checking the flag instead of doing  `===` comparison.

I think the next if statement below might also have the same issue, but I don't have touch controls to test with.


yomotsu:
Sorry for the delay...
Thanks for pointing that out. You are right!
Let me merge this first, and I'll fix others too.

PR #343, opened 2023-01-25
missing `EventDispatcher`.`hasEventListener` method

abernier:
I got the following TS error, while adding a new `makeDefault` prop for the drei component in https://github.com/pmndrs/drei/pull/1247:

image

The `EventDispatcher.hasEventListener` was commented, I'm not sure why... I tried uncommented it and check the tests were still ok.

I let you confirm @yomotsu

### DIFF
--- a/src/EventDispatcher.ts
+++ b/src/EventDispatcher.ts
@@ -25,13 +25,19 @@ export class EventDispatcher {
 
 	}
 
-	// hasEventListener( type: string, listener: Listener ): boolean {
+	/**
+	 * Presence of the specified event listener.
+	 * @param type event name
+	 * @param listener handler function
+	 * @category Methods
+	 */
+	hasEventListener( type: string, listener: Listener ): boolean {
 
-	// 	const listeners = this._listeners;
+		const listeners = this._listeners;
 
-	// 	return listeners[ type ] !== undefined && listeners[ type ].indexOf( listener ) !== - 1;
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your report.
I didn't notice it...
Actually, I just would like the event system to be minimized since `hasEventListener` is not really useful so far I think.

Also, the d.ts doesn't expose `hasEventListener`, according to the code:
https://github.com/yomotsu/camera-controls/blob/dev/dist/EventDispatcher.d.ts

Maybe we could just remove the commented-out lines...?

yomotsu:
or probably it comes from the three.js's d.ts...somehow?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1e01ba158bd6e7aafc3fd6376364034f7dcb836b/types/three/src/core/EventDispatcher.d.ts#L37

abernier:
I'm not so sure, but I think it is required by fiber, which declares `controls` state as a `THREE.EventDispatcher`:

https://github.com/pmndrs/react-three-fiber/blob/8bab743dfeb686524c475f7d3135a6c48a371f1f/packages/fiber/src/core/store.ts#L105

I think the `hasEventListener` is then "required" to conform that interface.

abernier:
a question I have is why CC implements that `EventDispatcher` rather than picking it from `THREE`? https://github.com/mrdoob/three.js/blob/master/src/core/EventDispatcher.js

yomotsu:
@abernier I've published v1.38.1 with `hasEventListener()`

PR #350, opened 2023-01-28
updating r3f example

abernier:
could be useful for those coming from `camera-controls` and looking for r3f "official" example ;)


yomotsu:
Thank you for the change. I forgot that...

> could be useful

Exactly!
Could we have both links? like "r3f drei official(which you added in this PR)" and "simplest(old link)"

abernier:
Good idea! :)

I revert it in https://github.com/yomotsu/camera-controls/pull/350/commits/0e49ac85a5204c03b5b1d52c80a07c3ca3bb939d

yomotsu:
Thanks! just merged!

PR #355, opened 2023-02-03
Exposing `CameraControls.createBoundingSphere`

abernier:
# What

I found that it could be useful to borrow the inner `createBoundingSphere` function.
Starting with the `examples/fit-to-bounding-sphere.html` which was needing it and has to redefined it.

=> I've just made this inner function a static method, and make `examples/fit-to-bounding-sphere.html` using it :)

# Why

Let me give you some context...

In my project, for debug purposes, I wanted to show the bounding sphere of `myMesh` (like in the `fit-to-bounding-sphere`). Primarily, I've created my own BS from `myMesh` but I figured out it was a little bigger that the one CC internally relies on to fitTo...

### DIFF
--- a/examples/fit-to-bounding-sphere.html
+++ b/examples/fit-to-bounding-sphere.html
@@ -64,7 +64,7 @@ scene.add( gridHelper );
 
 new GLTFLoader().load( './rubber-duck.glb', function ( gltf ) {
 
-	const boundingSphere = computeBoundingSphere( gltf.scene );
+	const boundingSphere = CameraControls.createBoundingSphere( gltf.scene );
 	const sphereHelper = new THREE.Mesh(
 		new THREE.SphereBufferGeometry( boundingSphere.radius, 16, 16 ),
 		new THREE.MeshBasicMaterial( { color: 0xff0000, wireframe: true } )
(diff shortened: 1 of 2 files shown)


src/CameraControls.ts
```diff
@@ -2946,64 +2946,64 @@ export class CameraControls extends EventDispatcher {
...
-		}
+		} );
 
-	} );
+		boundingSphere.radius = Math.sqrt( maxRadiusSq );
+		return boundingSphere;
 
-	boundingSphere.radius = Math.sqrt( maxRadiusSq );
-	return boundingSphere;
+	}
```
abernier, line 3007: all this is pure re-indentation: haven't touch the code

src/CameraControls.ts
```diff
@@ -2946,64 +2946,64 @@ export class CameraControls extends EventDispatcher {
 
 	}
 
-}
+	static createBoundingSphere( object3d: _THREE.Object3D, out: _THREE.Sphere = new THREE.Sphere() ): _THREE.Sphere {
```
abernier, line 2949: Here, I've set up a default for `out`: `= new THREE.Sphere()`

examples/fit-to-bounding-sphere.html
```diff
@@ -64,7 +64,7 @@
 
 new GLTFLoader().load( './rubber-duck.glb', function ( gltf ) {
 
-	const boundingSphere = computeBoundingSphere( gltf.scene );
+	const boundingSphere = CameraControls.createBoundingSphere( gltf.scene );
```
abernier, line 67: re-using the static method here

src/CameraControls.ts
yomotsu: thanks for the comment. helpful😊

src/CameraControls.ts
yomotsu: okay👍

yomotsu:
Thanks for your PR as usual.
I understand the background and your use case.
LGTM. Let me merge this.

PR #358, opened 2023-02-03
fix(examples/fitToSphere): reset button

abernier:
`reset` button was currently broken due to `cameraControls` var is not global due to `<script type="module">`

This PR just fix this.


yomotsu:
Oops, right. Thanks for fixing it.

PR #365, opened 2023-02-10
semantic-release

abernier:
- [x] localhost Verdaccio publish
  - [x] @yomotsu has reproduced
  - [x] test the resulting package in a working/existing project:
    ```sh
    $ npm i camera-controls --registry=http://localhost:4873
    ```
- [x] Test the release `.github/workflow` in [Actions](https://github.com/yomotsu/camera-controls/actions) -- on Github Packages registry (not to conflict with npm registry)
  - [x] @yomotsu has reproduced
  - [x] test the resulting package in a working/existing project:
    ```sh
    $ npm i @yomotsu/camera-controls --registry=https://npm.pkg.github.com
    ```
- [x] Integrate doc generation to the github-action workflow
  - [x]  @yomotsu has ([nearly](https://github.com/yomotsu/camera-controls/pull/365#user-content-fn-1-07f8c4865107bc0670127ebbcdfe44e3))reproduced
  
- [x] Once everything is tested, change some values (cf. comments) and delete tests-related files
  - [ ] @yomotsu final review

---

# Verdaccio (local registry)

To test it on your localhost:

1. Create a new local `semantic-release` branch from `dev`
   ```sh
   $ git checkout dev
   $ git checkout -b semantic-release
   ```
2. Pull this PR branch into:
   ```sh
   $ git remote add abernier https://github.com/abernier/camera-controls.git
   $ git pull abernier semantic-release
   ```
3. Push to a remote `yomotsu:semantic-release` branch:
```sh
   $ git push origin semantic-release # because a remote branch should exist for semantic-release to compare against
   ```
4. Launch a local npm registry:
   ```sh
   $ npx -y verdaccio --config tmp/verdaccio-config.yml
   ```
5. Create a Github [PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) with following permissions to the repo:
   - `contents: write`
   - `issues: write`
   - `pull-requests: write`
6. Try a release (on your local Verdaccio registry):
   ```sh
   $ export NPM_TOKEN=1234 # whatever, it is fake
   $ export GITHUB_TOKEN=github_pat_XXXXXXX # not fake this one!
   
   $ npm run release -- --ci false
   
   > camera-controls@2.2.1 release
   > semantic-release --repository-url abernier/camera-controls --ci false
   
   [7:30:07 PM] [semantic-release] › ℹ  Running semantic-release version 20.1.0
[...]

PR #377, opened 2023-02-19
`master` as release-branch

abernier:
In order to trigger a release when pushing on `master` (rather than `dev`) as discussed in https://github.com/yomotsu/camera-controls/pull/365

---

don't miss out to:
- [x] add the `master` branch in [Settings > Environments](https://github.com/yomotsu/camera-controls/settings/environments) > github pages

All the rest should be good if I don't forget anything...

NB: maybe if one day, merge-queue becomes available for personal repos, we could go back to `dev` ;)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Build and Release to npm
 on:
   push:
     branches:
-      - 'dev'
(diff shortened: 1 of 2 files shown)


yomotsu:
many thanks!
just merged

PR #378, opened 2023-02-19
Add PointerLock API compatibility

mjurczyk:
Add simple internal compatibility with [PointerLock API](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestPointerLock) / [three.js PointerLockControls](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/PointerLockControls.js).

Currently CameraControls acknowledges PointerLock API in `onPointerMove` by calculating [delta values](https://github.com/yomotsu/camera-controls/blob/dev/src/CameraControls.ts#L993) - but since controls become reactive only after at least one `onPointerDown / onMouseDown` event - PointerLock's arguably most practical use-case, a first-person camera, is not achievable (user would need to hold down a button to make controls listen to events.)

Changes:
* `pointerlockchange` and `pointerlockerror` listeners ([source](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/controls/PointerLockControls.js#L81))
* automatically listen to pointer movement when pointer lock is active (since `.requestPointerLock` doesn't require pointer event, also mock a pointer object internally if needed)
* add `pointer-lock` example (the only thing necessary to engage pointer-lock controls is to use `document.requestPointerLock()`, no changes to CameraControls API)

https://user-images.githubusercontent.com/9549760/219977456-7f5298f3-2e4c-46bc-98c5-6fd859109a27.mov

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -500,6 +500,21 @@ export class CameraControls extends EventDispatcher {
 		const lastDragPosition = new THREE.Vector2() as _THREE.Vector2;
 		const dollyStart = new THREE.Vector2() as _THREE.Vector2;
 
+		const disposePointer = ( pointer?: PointerInput ) => {
+
+			const isPointerLockActive = this._domElement && this._domElement.ownerDocument.pointerLockElement === this._domElement;
+
+			if ( ! pointer || isPointerLockActive && this._activePointers.length <= 1 ) {
+
+				// NOTE Do not dispose pointer-locked pointer
+				return;
+
+			}
+
+			this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
+
+		};
+
 		const onPointerDown = ( event: PointerEvent ) => {
 
 			if ( ! this._enabled || ! this._domElement ) return;
@@ -516,8 +531,7 @@ export class CameraControls extends EventDispatcher {
 
 			if ( mouseButton !== null ) {
 
-				const zombiePointer = this._findPointerByMouseButton( mouseButton );
-				zombiePointer && this._activePointers.splice( this._activePointers.indexOf( zombiePointer ), 1 );
+				disposePointer( this._findPointerByMouseButton( mouseButton ) );
 
 			}
 
@@ -554,8 +568,7 @@ export class CameraControls extends EventDispatcher {
 
 			if ( mouseButton !== null ) {
 
-				const zombiePointer = this._findPointerByMouseButton( mouseButton );
-				zombiePointer && this._activePointers.splice( this._activePointers.indexOf( zombiePointer ), 1 );
+				disposePointer( this._findPointerByMouseButton( mouseButton ) );
 
 			}
 
@@ -685,15 +698,7 @@ export class CameraControls extends EventDispatcher {
 
 		const onPointerUp = ( event: PointerEvent ) => {
 
-			const pointerId = event.pointerId;
-			const isPointerLockActive = this._domElement && this._domElement.ownerDocument.pointerLockElement === this._domElement;
-
-			if ( ! isPointerLockActive || ( isPointerLockActive && this._activePointers.length > 1 ) ) {
-
-				const pointer = this._findPointerById( pointerId );
-				pointer && this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
-
-			}
+			disposePointer( this._findPointerById( event.pointerId ) );
 
 			if ( event.pointerType === 'touch' ) {
 
@@ -733,8 +738,7 @@ export class CameraControls extends EventDispatcher {
 
 		const onMouseUp = () => {
 
-			const pointer = this._findPointerById( 0 );
-			pointer && this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
+			disposePointer( this._findPointerById( 0 ) );
 			this._state = ACTION.NONE;
 
 			endDragging();
@@ -830,8 +834,7 @@ export class CameraControls extends EventDispatcher {
 					event instanceof MouseEvent ? 0 :
 					0;
 
-				const pointer = this._findPointerById( pointerId );
-				pointer && this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
+				disposePointer( this._findPointerById( pointerId ) );
 
 				// eslint-disable-next-line no-undef
 				this._domElement.ownerDocument.removeEventListener( 'pointermove', onPointerMove, { passive: false } as AddEventListenerOptions );
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -686,8 +686,14 @@ export class CameraControls extends EventDispatcher {
 		const onPointerUp = ( event: PointerEvent ) => {
 
 			const pointerId = event.pointerId;
-			const pointer = this._findPointerById( pointerId );
-			pointer && this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
+			const isPointerLockActive = this._domElement && this._domElement.ownerDocument.pointerLockElement === this._domElement;
+
+			if ( ! isPointerLockActive || ( isPointerLockActive && this._activePointers.length > 1 ) ) {
+
+				const pointer = this._findPointerById( pointerId );
+				pointer && this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
+
+			}
 
 			if ( event.pointerType === 'touch' ) {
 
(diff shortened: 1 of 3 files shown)


src/CameraControls.ts
```diff
@@ -1117,6 +1122,58 @@ export class CameraControls extends EventDispatcher {
...
+			// Element.requestPointerLock is allowed to happen without any pointer active - create a faux one for compatibility with controls
+			const pointer: PointerInput = {
+				pointerId: 1,
+				clientX: 0,
+				clientY: 0,
+				deltaX: 0,
+				deltaY: 0,
+				mouseButton: null
+			};
+			this._activePointers.push( pointer );
```
yomotsu, line 1138: Shouldn't there be only one pointer during pointer lock?
If so, we can call `this.cancel()` to remove all this._activePointers before this line.

mjurczyk: Yes! Also - one bug I noticed, when pointer is locked and you click (ie. trigger one of the onPointer events), all the `dragging` chain is executed - and the last step of it is to remove `pointermove` listeners.

Is there a clean way we can avoid that - or would just checking for pointer lock make sense?

yomotsu: Ah right.
I think there is no clean way for this particular case.
So, I would do:

> would just checking for pointer lock make sense?

mjurczyk:
@yomotsu added cancel call where pointer lock ends as well. Moved `_activePointers.splice` to a separate method to reduce duplication a bit - that way code for checking pointer lock can be in just one place.

src/CameraControls.ts
```diff
@@ -500,6 +500,21 @@ export class CameraControls extends EventDispatcher {
 		const lastDragPosition = new THREE.Vector2() as _THREE.Vector2;
 		const dollyStart = new THREE.Vector2() as _THREE.Vector2;
 
+		const disposePointer = ( pointer?: PointerInput ) => {
```
yomotsu, line 503: Can `pointer` be a required argument?
I would prefer like:
```ts
const pointer = ...
pointer && disposePointer( pointer );
```

mjurczyk: Done

mjurczyk: Of course - done 😁

yomotsu:
Thanks for the changes.
Let me discuss a little bit.

This PR enables auto mouse dragging (which is rotation) when the pointer is locked.
But I think some ppl may not prefer the auto-mouse-dragging behavior.
like even pointer is locked they might prefer mouse dragging without a visible pointer.

So, what do you think if we have a boolean property that enables auto-mouse-dragging?

src/CameraControls.ts
```diff
@@ -1002,7 +1021,8 @@ export class CameraControls extends EventDispatcher {
 				( this._state & ACTION.ROTATE ) === ACTION.ROTATE ||
 				( this._state & ACTION.TOUCH_ROTATE ) === ACTION.TOUCH_ROTATE ||
 				( this._state & ACTION.TOUCH_DOLLY_ROTATE ) === ACTION.TOUCH_DOLLY_ROTATE ||
-				( this._state & ACTION.TOUCH_ZOOM_ROTATE ) === ACTION.TOUCH_ZOOM_ROTATE
+				( this._state & ACTION.TOUCH_ZOOM_ROTATE ) === ACTION.TOUCH_ZOOM_ROTATE ||
+				isPointerLockActive
```
yomotsu, line 1025: How about applying the mouse left button setting for pointerLock?
(If the pointer is locked, it assumes the state as the left button action.

```
			const state = isPointerLockActive ? this.mouseButtons.left : this._state;

			if (
				( state & ACTION.ROTATE ) === ACTION.ROTATE ||
				( state & ACTION.TOUCH_ROTATE ) === ACTION.TOUCH_ROTATE ||
				( state & ACTION.TOUCH_DOLLY_ROTATE ) === ACTION.TOUCH_DOLLY_ROTATE ||
				( state & ACTION.TOUCH_ZOOM_ROTATE ) === ACTION.TOUCH_ZOOM_ROTATE
			) {
```

then do the same for the rest conditions as well.

src/CameraControls.ts
```diff
@@ -500,6 +500,21 @@ export class CameraControls extends EventDispatcher {
...
+		const disposePointer = ( pointer: PointerInput ) => {
+
+			const isPointerLockActive = this._domElement && this._domElement.ownerDocument.pointerLockElement === this._domElement;
+
+			if ( isPointerLockActive && this._activePointers.length <= 1 ) {
+
+				// NOTE Do not dispose pointer-locked pointer
+				return;
+
+			}
```
yomotsu, line 512: Would we need this? It seems this does not happen.

src/CameraControls.ts
```diff
@@ -1117,6 +1137,60 @@ export class CameraControls extends EventDispatcher {
...
 
+		const lockPointer = (): void => {
+
+			if ( ! this._enabled || ! this._domElement ) return;
+
+			this.cancel();
+
+			// Element.requestPointerLock is allowed to happen without any pointer active - create a faux one for compatibility with controls
+			const pointer: PointerInput = {
+				pointerId: 1,
```
yomotsu, line 1148: maybe we should use a unique id for pointer-lock exclusively.
```suggestion
				pointerId: - 1,
```

mjurczyk: Pointer id of the pointer locked pointer is `1` (since there's only one) - not sure if it's modifiable 👀

src/CameraControls.ts
mjurczyk: That's the check mentioned earlier - when you lock pointer, then press mouse1 a few times and drag around, the `endDragging` will dispose `pointermove` event listener and make pointerlock not work again.

src/CameraControls.ts
mjurczyk: Any downsides to this? I can imagine one scenario, where pointer locked app would like to pan with mouse1 pressed - and they couldn't because mouse1 would be "always on" when pointer is locked.

mjurczyk:
> This PR enables auto mouse dragging (which is rotation) when the pointer is locked.
But I think some ppl may not prefer the auto-mouse-dragging behavior.
like even pointer is locked they might prefer mouse dragging without a visible pointer.
> 
> So, what do you think if we have a boolean property that enables auto-mouse-dragging?
> 
> So how about exposing lockPointer() method to enable auto mouse dragging?

Yup, I agree - the API proposed in PR right now is not perfect. We can add a new method - or to solve [this one](https://github.com/yomotsu/camera-controls/pull/378#discussion_r1115704206) at the same time and keep stuff consistent - we could add a similar config as `cameraConfig.mouseButtons`, with `cameraConfig.pointerMove` ? 🤔

src/CameraControls.ts
yomotsu: Ah, okay I checked other pointer entries, and it seems so.
let me mark this comment as resolved.

src/CameraControls.ts
yomotsu: sorry for the lack of information.
I meant if we expose the method, we can set `true` to the flag.

```ts
lockPointer () {
  this._pointerLocked = true;
  // ...snip
  // - call requestPointerLock
  // - enable auto dragging
}
```

then if the pointer lock is called via the method, we will apply the mouse left action.

```diff
- const state = isPointerLockActive ? this.mouseButtons.left : this._state;
+ const state = this._pointerLocked ? this.mouseButtons.left : this._state;
```

If users call `requestPointerLock` in user-land directly, `this._pointerLocked` won't be `true` and they can use original actions.

(Sorry if I was misunderstanding...

src/CameraControls.ts
yomotsu: I understand.
Could we check the pointer before calling disposePointer?

mjurczyk: Kinda defeats the purpose of disposePointer 🥲

yomotsu: I'm sorry, no offense meaning...🥲
I thought, `disposePointer()` should forcibly dispose of the pointer in any case.
And now `this._activePointers.splice(...)` has a name as `disposePointer()` which is better readability than mine...

maybe we could change the name of the function if it doesn't dispose of in some cases.

mjurczyk: No worries - would you like to push the changes yourself? I just wanted to add a suggestion to handle pointer lock - all the stylistic changes in the code I can leave up to you - you can push right to this PR and we merge 😌

yomotsu: @mjurczyk sorry for the delay, and thanks for understanding. Okay, I will make a PR to yours!

yomotsu:
@mjurczyk 
Thanks for the PR.
As you mentioned, I will merge this to `feature/pointer-clock` branch first, then I will add some chnages.

PR #383, opened 2023-02-25
esm examples scripts

abernier:
a single `<script type="module">import ...` for `examples/*`


yomotsu:
Thanks!
I've been thinking of doing that too!
meged.

PR #387, opened 2023-03-08
expose lockPointer and unlockPointer methods

yomotsu:
- `lockPointer()` and `unlockPointer()` are now public methods.
- Previously, the pointer created with PointerLock was an ordinary pointer and could be accidentally removed, but I added the `this._lockedPointer` field as an instance of the special pointer, to save from the problem.
- accept the right button dragging

https://user-images.githubusercontent.com/212837/223721937-89f6680c-4427-4fa6-a723-52b82c98d10e.mp4

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -504,12 +504,6 @@ export class CameraControls extends EventDispatcher {
 		const lastDragPosition = new THREE.Vector2() as _THREE.Vector2;
 		const dollyStart = new THREE.Vector2() as _THREE.Vector2;
 
-		const disposePointer = ( pointer: PointerInput ) => {
-
-			this._activePointers.splice( this._activePointers.indexOf( pointer ), 1 );
-
-		};
-
 		const onPointerDown = ( event: PointerEvent ) => {
 
 			if ( ! this._enabled || ! this._domElement ) return;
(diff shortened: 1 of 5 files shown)


src/CameraControls.ts
```diff
@@ -532,11 +527,13 @@ export class CameraControls extends EventDispatcher {
 
 			if ( mouseButton !== null ) {
 
-				const pointer = this._findPointerByMouseButton( mouseButton );
-				pointer && disposePointer( pointer );
+				const zombiePointer = this._findPointerByMouseButton( mouseButton );
+				zombiePointer && disposePointer( zombiePointer );
```
yomotsu, line 531: mark the pointer as completed execution with the name as "zombie".

src/CameraControls.ts
```diff
@@ -1138,57 +1157,67 @@ export class CameraControls extends EventDispatcher {
...
-				pointerId: 1,
+			this._lockedPointer = {
+				pointerId: - 1,
 				clientX: 0,
 				clientY: 0,
 				deltaX: 0,
 				deltaY: 0,
-				mouseButton: null
+				mouseButton: null,
 			};
```
yomotsu, line 1174: save as `this._lockedPointer` while pointer-lock is active.
then
- `this._lockedPointer` should not be removed from `activePointers`
- if `this._lockedPointer` is not `null`, that means pointer-lock is active.

yomotsu:
@mjurczyk
I've made some changes from your PR. What do you think about this change...?

PR #392, opened 2023-03-14
fix: position in fromJSON does not work correctly

yomotsu:


src/CameraControls.ts
```diff
@@ -2625,7 +2624,7 @@ export class CameraControls extends EventDispatcher {
 		this._focalOffset0.fromArray( obj.focalOffset0 );
 
 		this.moveTo( obj.target[ 0 ], obj.target[ 1 ], obj.target[ 2 ], enableTransition );
-		_sphericalA.setFromVector3( position.sub( this._targetEnd ).applyQuaternion( this._yAxisUpSpace ) );
+		_sphericalA.setFromVector3( _v3A.fromArray( obj.position ).sub( this._targetEnd ).applyQuaternion( this._yAxisUpSpace ) );
```
yomotsu, line 2627: `position`  which is `_v3A`  was overwritten by the `moveTo()` method and caused the problem.

PR #408, opened 2023-05-01
chore(release): actions/setup-node reads .node-version file

abernier:
so there is no need to hard-code it into the workflow file: https://stackoverflow.com/a/62978089/133327


yomotsu:
Thanks for your PR as usual!
just meged

PR #412, opened 2023-05-15
fix: verticalDragToForward has no effect when used with OrthographicCamera

### LINKED ISSUE #411: verticalDragToForward does not work with OrthographicCamera
### Describe the bug

`verticalDragToForward` option set to true does not have an effect when camera-controls are used with `THREE.OrthographicCamera`. There is no difference between true/false values. 

This results in slower dragging of the camera in the direction in which it is angled towards, when it is limited by a bounding box (`setBoundary`)

### To Reproduce

1. Go to "boundary" example from the repo https://yomotsu.github.io/camera-controls/examples/boundary.html
2. Add boundary to the scene
3. Compare behaviour of the camera with and without "vertical drag to move forward" checked (try slamming the yellow centre point agains the "ceiling" of the green cube)
4. Try to do the same, but with `OrthographicCamera` - see https://codesandbox.io/s/camera-controls-issue-ptxvb6?file=/src/index.js 

### Code

_No response_

### Live example

https://codesandbox.io/s/camera-controls-issue-ptxvb6?file=/src/index.js

### Expected behavior

Given:
- CameraControls are used with OrthographicCamera
- CameraControls.verticalDragToForward = true
- CameraControls has a bounding box set

When:
 - camera target reaches the edge of boundary and can no longer move in the "up" direction

Then:
- camera instead moves in the "forwards" direction (like in the example with `PerspectiveCamera`)

### Screenshots or Video

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

mixvar:
closes #411

`if ( this.verticalDragToForward )` check was nested inside `if ( isPerspectiveCamera( this._camera ) )` and not repeated inside `else if ( isOrthographicCamera( this._camera ) )`. I am assuming this was not intentional. 

Now the code for applying `_truckInternal` is the same for both camera types, except `truckX` and `pedestalY` are resolved in different ways.

I changed camera to orthographic in some examples and it seems to work fine with `verticalDragToForward` enabled but more testing is needed I think.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -2937,47 +2937,54 @@ export class CameraControls extends EventDispatcher {
 
 	protected _truckInternal = ( deltaX: number, deltaY: number, dragToOffset: boolean ): void => {
 
+		let truckX: number;
+		let pedestalY: number;
+
 		if ( isPerspectiveCamera( this._camera ) ) {
 
 			const offset = _v3A.copy( this._camera.position ).sub( this._target );
 			// half of the fov is center to top of screen
 			const fov = this._camera.getEffectiveFOV() * DEG2RAD;
 			const targetDistance = offset.length() * Math.tan( fov * 0.5 );
-			const truckX    = ( this.truckSpeed * deltaX * targetDistance / this._elementRect.height );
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your PR.
Looks good to me!

yomotsu:
merged. many thanks!

Issue #417, opened 2023-05-26
Using `enableTransition=true` in dollyTo results in stuttery motion when `dollyToCursor` is on


hybridherbst:
### Describe the bug

Related to
- #230.

When doing as suggested there and calling dollyTo with enableTransition=true, while the dolly becomes smooth now the camera pan jitters around.

With `enableTransition=false`:

https://github.com/yomotsu/camera-controls/assets/2693840/85efdec3-1800-4f78-bcdb-3d1dc46eec97

With `enableTransition=true`:

https://github.com/yomotsu/camera-controls/assets/2693840/f8b0ca65-52c4-481d-a489-046602a79bfc

### To Reproduce

Change .dollyTo( distance ) to .dollyTo( distance, true) in
https://github.com/yomotsu/camera-controls/blob/22e2b6a09e8b48bcad9f5dcc79b93ef57e54b5aa/src/CameraControls.ts#L2182-L2215

and zoom into something.

### Code

```js
this._orbit.dampingFactor = 0.05;
this._orbit.dollyToCursor = true;

+ change as outlined above
```

### Expected behavior

Smooth behaviour

### Screenshots or Video

See above

### Device

Desktop

### OS

Windows

### Browser

Chrome

yomotsu:
Thanks for using camera-controls.
The reported problem is reproduced.

Although changing internal methods is not guaranteed, I think I can fix it.

yomotsu:
Then we could change to `.dollyTo( distance, true)` in `_dollyInternal` by default after #418 is merged as well.

yomotsu:
The reported problem should be fixed. Also, I've changed:

> Change .dollyTo( distance ) to .dollyTo( distance, true) in

By default. no need to change the internal method anymore.
Would you mind trying the latest version?

hybridherbst:
Will do in the next days! Thank you!

marwie:
Hello @yomotsu I just updated from 1.x to latest and it seems to be fixed. Thank you!

https://github.com/yomotsu/camera-controls/assets/5083203/45cbe35a-22d7-4820-a3ed-2973e7b29590


https://github.com/yomotsu/camera-controls/assets/5083203/c1c78b7a-d18f-436c-a14f-cebdb371583b

yomotsu:
Good to hear that, and thank you for sharing the videos

marwie:
Thank you for the library! The videos were taken from https://songsofcultures.com btw 😇

PR #418, opened 2023-05-27
fix: dollyToCursor with smooth

yomotsu:
This PR is to fix https://github.com/yomotsu/camera-controls/issues/417


yomotsu:
(let me test a little bit more...

PR #448, opened 2023-08-06
chore: Treeshake Docs add 'MathUtils'

vis-prime:
scroll uses Mathutils , so adding it to Tree shake docs

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -119,6 +119,7 @@ const subsetOfTHREE = {
 	Box3      : Box3,
 	Sphere    : Sphere,
 	Raycaster : Raycaster,
+	MathUtils : MathUtils,
 };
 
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your PR!
You mean here, right?
https://github.com/yomotsu/camera-controls/blob/cee042753169f3bbeb593833ce92d70d52b6862f/src/CameraControls.ts#L3203

Actually, that has to be:
```js
const clampedDistance = clamp( distance, this.minDistance, this.maxDistance );
```
(with internal `clamp` function)
Then no need to import three's MathUtils.

Would it be possible to make a PR for it instead of this PR...?

vis-prime:
Will do !

yomotsu:
Thanks!

PR #453, opened 2023-08-18
Re-normalize new camera up after cross product

a3ng7n:
After running `applyCameraUp`, the `_camera.up` vector gets used to create a quaternion in `updateCameraUp` via `setFromUnitVectors` - since the cross product of two vectors isn't normalized unless the two vectors are orthogonal two eachother, the resulting `_camera.up` ends up being non-normal, and causes the quaternion generation to get upset. This change just re-normalizes the output of the cross product.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -2610,7 +2610,7 @@ export class CameraControls extends EventDispatcher {
 		const side = _v3B.crossVectors( cameraDirection, this._camera.up );
 		// Then find the vector orthogonal to both this "side" vector and the "view" vector.
 		// This vector will be the new "up" vector.
-		this._camera.up.crossVectors( side, cameraDirection );
+		this._camera.up = _v3C.crossVectors( side, cameraDirection ).normalize();
 		this._camera.updateMatrixWorld();
 
 		const position = this.getPosition( _v3A );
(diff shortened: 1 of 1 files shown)


src/CameraControls.ts
```diff
@@ -2610,7 +2610,7 @@ export class CameraControls extends EventDispatcher {
 		const side = _v3B.crossVectors( cameraDirection, this._camera.up );
 		// Then find the vector orthogonal to both this "side" vector and the "view" vector.
 		// This vector will be the new "up" vector.
-		this._camera.up.crossVectors( side, cameraDirection );
+		this._camera.up = _v3C.crossVectors( side, cameraDirection ).normalize();
```
yomotsu, line 2613: Could this be:
```suggestion
		this._camera.up.crossVectors( side, cameraDirection ).normalize();
```
`crossVectors` returns `this`, which is the current vector3.

yomotsu:
@a3ng7n 
Thanks for your PR. That makes sense!
Would you mind checking the suggestion above?

src/CameraControls.ts
a3ng7n: Yup - everything seems to work with that. I'll update the PR.

a3ng7n:
> @a3ng7n Thanks for your PR. That makes sense! Would you mind checking the suggestion above?

Anytime! @yomotsu accepted the suggestion - didn't realize that's how that worked.

yomotsu:
Thanks for the change.
I just merged and will publish a patch updated version later

PR #454, opened 2023-08-29
fix: deprecated geometry from examples

seantai:
BoxBufferGeometry -> BoxGeometry
SphereBufferGeometry -> SphereGeometry

I found this after looking at the fit-to-bounding-sphere example and got the error:
Uncaught (in promise) TypeError: THREE.SphereBufferGeometry is not a constructor
    at fit-to-bounding-sphere.html:69:3


yomotsu:
Thanks for pointing out. That's right!
merged!

Issue #458, opened 2023-09-21
Add back modifier keys to pointer events


jamesnakagawa:
### Is your feature request related to a problem? Please describe.

At my company, we are having an issue with using modifier keys like in the example given. The most common example is when the user cmd+tabs away from the window so that the window loses focus. In this case the `keydown` event is registered but not `keyup`, so when the user switches back to the window, the modifier key is still thought to be active. We also implemented a listener on window focus, but still ran into edge cases

### Describe the solution you'd like

I would like if camera-controls handled modifier keys (alt, ctrl, and shift) in its pointer event handler similar to its previous behaviour.

### Describe alternatives you've considered

Using keydown/keyup like in the example
Using window focus event in order to mitigate the downsides

### Additional context

Thank you for this excellent library!
If I can offer to PR this change, I would be happy to.

yomotsu:
Thanks for using camera-controls.

> If I can offer to PR this change, I would be happy to.

PRs and issues are open to anyone. I would be happy to have PRs too.

jamesnakagawa:
@yomotsu 

Sorry to revisit old discussions.

Are you open to the idea still? I can create a new PR according to your suggestions.

ちなみに、日本に到着しました！

jamesnakagawa:
Oops, I pressed the wrong button. Please re-open if you are able to.

PR #459, opened 2023-09-23
Change mouse button config to a function `(e: Event) => ACTION`

jamesnakagawa:
Regarding #458,

I propose a new structure for configuring mouse buttons. Instead of a configuration object, user can pass a configuration _function_ which receives an `Event` (either wheel event, click event, or contextmenu event) and which returns the intended action they want. With this, user can access `event.shiftKey` and all other modifier key flags and reliably set them without overly verbose configuration.

I have also updated the example at `examples/mouse-drag-with-modifier-keys.html` to show how it could be used.

Lastly, I wrote a custom setter for the previous configuration, `mouseButtons`, so that this is not a breaking change.

I'm not familiar with the coding conventions of this project, so please excuse if I didn't document something correctly. I look forward to hearing your feedback on my proposal!

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -314,23 +314,20 @@ export class CameraControls extends EventDispatcher {
 	// button configs
 	/**
 	 * User's mouse input config.
+	 * @param MouseEvent or undefined. Can be a click event, wheel event, or contextmenu event.
+	 * @returns CameraControls.ACTION (ROTATE, TRUCK, OFFSET, DOLLY, ZOOM, NONE)
 	 *
-	 * | button to assign      | behavior |
-	 * | --------------------- | -------- |
-	 * | `mouseButtons.left`   | `CameraControls.ACTION.ROTATE`* \| `CameraControls.ACTION.TRUCK` \| `CameraControls.ACTION.OFFSET` \| `CameraControls.ACTION.DOLLY` \| `CameraControls.ACTION.ZOOM` \| `CameraControls.ACTION.NONE` |
-	 * | `mouseButtons.right`  | `CameraControls.ACTION.ROTATE` \| `CameraControls.ACTION.TRUCK`* \| `CameraControls.ACTION.OFFSET` \| `CameraControls.ACTION.DOLLY` \| `CameraControls.ACTION.ZOOM` \| `CameraControls.ACTION.NONE` |
-	 * | `mouseButtons.wheel` ¹ | `CameraControls.ACTION.ROTATE` \| `CameraControls.ACTION.TRUCK` \| `CameraControls.ACTION.OFFSET` \| `CameraControls.ACTION.DOLLY` \| `CameraControls.ACTION.ZOOM` \| `CameraControls.ACTION.NONE` |
-	 * | `mouseButtons.middle` ² | `CameraControls.ACTION.ROTATE` \| `CameraControls.ACTION.TRUCK` \| `CameraControls.ACTION.OFFSET` \| `CameraControls.ACTION.DOLLY`* \| `CameraControls.ACTION.ZOOM` \| `CameraControls.ACTION.NONE` |
-	 *
-	 * 1. Mouse wheel event for scroll "up/down" on mac "up/down/left/right"
-	 * 2. Mouse click on wheel event "button"
-	 * - \* is the default.
-	 * - The default of `mouseButtons.wheel` is:
+	 * - The default of `WheelEvent` is:
 	 *   - `DOLLY` for Perspective camera.
 	 *   - `ZOOM` for Orthographic camera, and can't set `DOLLY`.
-	 * @category Properties
+	 * - The default for other `MouseEvent`s is:
+	 *   - `ROTATE` if left button is pressed, otherwise
+	 *   - `ZOOM` if middle button is pressed, otherwise
+	 *   - `TRUCK` if right button is pressed
+	 * - The default is `ROTATE` if event is undefined
+	 * @category Methods
 	 */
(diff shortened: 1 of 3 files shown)


src/CameraControls.ts
```diff
@@ -714,22 +737,9 @@ export class CameraControls extends EventDispatcher {
...
+					this._isDragging && ( event.buttons & MOUSE_BUTTON.ALL ) > 0 ) {
 
-					this._state = this._state | this.mouseButtons.middle;
-
-				}
-
-				if ( this._isDragging && ( event.buttons & MOUSE_BUTTON.RIGHT ) === MOUSE_BUTTON.RIGHT ) {
-
-					this._state = this._state | this.mouseButtons.right;
+					this._state = this._state | this.mouseEventToAction( event );
```
jamesnakagawa, line 742: I was able to simplify this conditional because of this change.

`( event.buttons & MOUSE_BUTTON.ALL ) > 0` looks a little weird, and technically it would exclude `event.buttons === 8 || event.buttons === 16`, which is not a common use case. I wanted the result of the function to match previous behaviour exactly.

src/CameraControls.ts
```diff
@@ -1019,25 +1019,11 @@ export class CameraControls extends EventDispatcher {
...
-					this._state = this._state | this.mouseButtons.left;
-
-				}
-
-				if ( ( event.buttons & MOUSE_BUTTON.MIDDLE ) === MOUSE_BUTTON.MIDDLE ) {
-
-					this._state = this._state | this.mouseButtons.middle;
+			} else if ( ! this._lockedPointer && ( event.buttons & MOUSE_BUTTON.LEFT ) === MOUSE_BUTTON.LEFT ||
+				( event.buttons & MOUSE_BUTTON.MIDDLE ) === MOUSE_BUTTON.MIDDLE ||
+				( event.buttons & MOUSE_BUTTON.RIGHT ) === MOUSE_BUTTON.RIGHT ) {
```
jamesnakagawa, line 1024: I do not use `MOUSE_BUTTON.ALL` here because having `! this._lockedPointer &&` changes the logic.

jamesnakagawa:
@yomotsu Just wondering if you had a chance to see this?

yomotsu:
Sorry for the delay, I've been busy with [house moving](https://twitter.com/yomotsu/status/1706352166317986035).
And thank you for the PR.

This contains a lot change than I thought...and going to be a breaking change (no compatibility with older versions).
Let me confirm a little bit:

- How do we handle if the user presses with "A" key or others, like, rotate for the R key, translate for the T key, etc. This could be multiple keys at the same time. (Therefore, the demo code was handling additional keys without MouseEvents)

- Camera Controls accepts touch events via Pointer Events too. Can `mouseEventToAction` handle any pointer events?

jamesnakagawa:
引っ越しおめでとう！

Thank you for the response.

1) If I understand correctly, using "R" for rotate, "T" for transform, means that pressing once would change the mode but it does not need to be held down?

In that case, it could be done using closure:

```
let mode = 'rotate';

document.addEventListener('keypress', e => {
    if (e.key === 'R' || e.key === 'r') {
        mode = 'rotate';
    } else if (e.key === 'T' || e.key === 't') {
        mode = 'translate';
    }
});

cameraControls.mouseEventToAction = event => {
	if ( event instanceof MouseEvent ) {
		if ( ( event.buttons & MOUSE_BUTTON.LEFT ) === MOUSE_BUTTON.LEFT ) {
			if (mode === 'translate') {
				return ACTION.TRUCK;
			}
			if (mode === 'rotate') {
				return ACTION.ROTATE;
			}
			// etc
		}
	}
};
```

2) For `PointerEvent`, it subclasses `MouseEvent` and properly sets `event.buttons` and `event.shiftKey/metaKey/etc`, so I see no incompatibility here.

3) Re: breaking change. Now that I think about this more, you are right, you can't set deep values using this compatibility bridge and turning `mouseEventToAction` back into `mouseButtons` is quite complex. In any case, probably not worth it.

Could this wait for the next major version of camera-controls?

Otherwise, the alternative would be to extend `mouseButtons` like this:

```
this.mouseButtons = {
	left: ACTION.ROTATE,
	middle: ACTION.DOLLY,
	right: ACTION.TRUCK,
	wheel: ACTION.DOLLY,
	'left.ctrl': ACTION.TRUCK,
	'left.shift': ACTION.TRUCK,
	'left.meta': ACTION.DOLLY,
	'left.ctrl.shift': null,
	'left.ctrl.shift.meta': ACTION.DOLLY,
};
```

Where a `null`/`undefined` value would indicate to inherit from the base setting. i.e `'left.ctrl.shift': null` would inherit from `'left.ctrl': ACTION.TRUCK` instead.

This would be much more verbose, which is why I think you didn't want this feature in the first place.

jamesnakagawa:
@yomotsu Any thoughts on this? I'm happy to do another PR or make changes based on the direction you'd like to go.

yomotsu:
Sorry again for the delay.
Could this be a none break change PR?
I mean,

- It's okay to add the interrupting feature as an advanced usage.
- However, I also would like to keep the simple setting feature.

Also, what do you think about the following:

Renaming `MOUSE_BUTTON.ALL` to `MOUSE_BUTTON.ANY`.
Renaming `mouseEventToAction` to `pointerEventToAction` and can handle other pointer events such as touch and pen.

jamesnakagawa:
We could do it that way, yes. If `pointerEventToAction == null` or `undefined`, then use `mouseButtons`, otherwise `mouseButtons` will be ignored. Is that what you were thinking?

PR #466, opened 2023-10-23
Execute dragging() once per frame

wwwillw:
I was having an issue panning with two fingers. If I drag my fingers side by side (like riding inline skates) it pans as expected. But if I drag my fingers one in front of the other (like riding a skateboard) it would cause the screen to zoom, even though I'm not changing the distance between my fingers.

I figured out that the behavior is caused by calling "dragging()" multiple times in a single frame. The browser usually calls the pointer move event for each finger at around the same time. This means the position data for both fingers is only in-sync for the second call, the first call is using out-of-date pointer position for the other finger. 

My solution is to add a _dragNeedsUpdate flag to ensure that dragging() would only be called once per frame.

Also I had to add tslib to the dev dependencies in order to run 'npm run build', so that's here as well.


yomotsu:
Thanks for your PR.
That makes sense and looks good to me.
merged!

PR #478, opened 2024-01-30
Bugfix for incorrect "lookInDirectionOf" behaviour when camera orbit target is not at origin

wellcaffeinated:
Reproduction/demonstration of the bug: https://jsfiddle.net/wellcaffeinated/zdb05n2u/

Correctly works when orbit target (green box) is at origin. But when moved (elevate), the look direction is improperly calculated.

This fixes the problem with the vector math.


yomotsu:
Thanks for your report and PR.
Let me merge this!

PR #506, opened 2024-05-02
Edit: 'out' the prop of 'getPosition' is now optional

kimgh06:
Hello, 

I just edited a little of codes of 'CameraControls.ts' file, 
```ts
getPosition( out: _THREE.Vector3, receiveEndValue: boolean = true ): _THREE.Vector3 {...}
```
to
```ts
getPosition( out?: _THREE.Vector3, receiveEndValue: boolean = true ): _THREE.Vector3 {...}
```

The previous code returns 'out' and sets 'out' twice.

if I try to get the position on your purpose, I shall try this:

```ts
import { Vector3 } from 'three'
let camPosition: Vector3 = new Vector3();
camControlRef.current.getPosition(camPosition);
```

However, actually i try to get the position of a camera like this:

```ts
import { Vector3 } from 'three'
const camPosition: Vector3 = camControlRef.current.getPosition();
```
but this code works and return the position of a camera.

And It has a problem using 'pmndrs/drei' package depending on this package with TypeScript, It occur type error "Expected 1-2 arguments, but got 0".

So, I have an idea how about being it optional for using the function in one line.

I hope you read this as soon as possible and answer.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -351,7 +351,7 @@ export class CameraControls extends EventDispatcher {
 	 * @category Methods
 	 */
 	// cancel will be overwritten in the constructor.
-	cancel: () => void = () => {};
+	cancel: () => void = () => { };
 
 	/**
 	 * Still an experimental feature.
@@ -404,7 +404,7 @@ export class CameraControls extends EventDispatcher {
 	protected _changedZoom = 0;
 
 	// collisionTest uses nearPlane. ( PerspectiveCamera only )
-	protected _nearPlaneCorners: [ _THREE.Vector3, _THREE.Vector3, _THREE.Vector3, _THREE.Vector3 ];
+	protected _nearPlaneCorners: [_THREE.Vector3, _THREE.Vector3, _THREE.Vector3, _THREE.Vector3];
 
 	protected _hasRested = true;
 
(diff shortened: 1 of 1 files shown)


yomotsu:
Thank you for bringing this to my attention.
Your point is well understood. However, I would prefer to make `out` a required value, similar to the classes in three.js.
For example, see the following code: https://github.com/mrdoob/three.js/blob/a2e9ee8204b67f9dca79f48cf620a34a05aa8126/src/math/Box3.js#L116-L126.

Therefore, we could consider removing the line below:
```
const _out = !! out && out.isVector3 ? out : new THREE.Vector3() as _THREE.Vector3;
```
(This line was primarily included for backward compatibility)

What do you think?

kimgh06:
OK, I had a mistake to explain what happens.

There's a problem when using TypeScript.
I use 'Vector3' of '@react-three/fiber' that only using as a type for 'react-three-fiber'.
And, the 'Vector3' of '@react-three/fiber' is already using to send position of 'mesh' that can not accept 'Of three.js' in TypeScript.

However, to use the function, I have to use 'Vector3' of 'three.js'.
image

If I am on your purpose, I should make a 'Vector3' object in my code and put it to 'getPosition' function as 'out'.
But I would like to code it with TypeScript.

So, They conflict in my code.

yomotsu:
What do you think about using "rename-import" for our imports? This approach could help us avoid conflicts

```
import { Vector3 as ThreeVector3 } from 'three';
```

or type import
```
import { type Vector3 } from '@react-three/fiber';
```

kimgh06:
It's a good propose, but that's not what I am saying.

I just would like to use like this without a type error:
```tsx
const camPosition: Vector3 = camControlRef.current.getPosition(); // <- But now it occurs a type error of "Expected 1-2 arguments, but got 0." that means "out" and "receiveEndValue" which is optional
```

So It's my propose that I want to make "out" to be opional.

yomotsu:
I mean...
```
camControlRef.current.getPosition( camPosition );
```
or
```
const camPosition = camControlRef.current.getPosition( camPosition );
```

PR #507, opened 2024-05-05
fix: reset focal offsets on input start and setLookAt

### LINKED ISSUE #303: setOrbitPoint breaks setLookAt
### Describe the bug

Setting setOrbitPoint seems to change the behavior of setLookAt. I would expect setLookAt to be idempotent, which is to say, when  given identical parameters, it should move the camera to the same position and point it at the same target. Calling setOrbitPoint in between calls to setLookAt seems to break this assumption, as subsequent calls to setLookAt with the same parameters move the camera to a new position/target. 

What is the best way to think about the relationship between setOrbitPoint and setLookAt? Should the code integrating camera-controls keep track of the last set orbit point, and use this when calculating the values to pass into setLookAt? E.g. use the last value passed to setOrbitPoint as an offset to setLookAt?

Thank you for your time and attention to this.

### To Reproduce

1. Go to https://yomotsu.github.io/camera-controls/examples/basic.html
2. In developer console, execute `cameraControls.setLookAt(3,5,2,3,0,-3)`
3. Execute `cameraControls.setLookAt(3,5,2,3,0,-3)` again, and observe that the camera positions doesn't change.
4. Execute `cameraControls.setOrbitPoint( 0, 0, 0 );`
5. Execute `cameraControls.setLookAt(3,5,2,3,0,-3)` again, and observe that the camera positions now changes.
6. Repeat steps 4-5

### Code

_No response_

### Live example

_No response_

### Expected behavior

* Calls to setOrbitPoint should not affect setLookAt
* setLookAt should be idempotent

Thank you for your time and attention to this.

### Screenshots or Video

Screen Capture_select-area_20220907224516 mp4

### Device

Desktop

### OS

Linux

### Browser

Firefox

### LINKED ISSUE #316: Restore CameraPosition after setOrbitPoint dosn't work
### Describe the bug

I rotate the scene the mouse point, for this I set the current hitpoint of the raycaster at setOrbitPoint(). 

This also works perfectly.

Now I reset the camera occasionally to a saved position. These are determined with cameraControls.getPosition() and cameraControls.getTarget(). 

The resetting of this position does not work after I changed the OrbitPoint once. Even if I reset the OrbitPoint to (0,0,0), which I set as default position, the camera is not set to the saved position anymore. 

Saving the camera state as JSON is unfortunately not possible, because other controllers may be used. Which do not offer this function. 

### To Reproduce

1. Store position with cameraControls.getPosition() and target with cameraControls.getTarget() in variable. 

2. Move and rotate an object, setOrbitPoint() must be called once to change the OrbitPoint.

3. The previous saved position cannot be restored with cameraControls.setPosition(x,y,z) and cameraControls.setTarget(x,y,z,). 

4. The saved position matches the current one, but the image is not the same anymore.

### Code

```js
//save the current camera status
var position = cameraControls.getPosition() 
var target = cameraControls.getTarget() 

--> Move and rotate object

//restore camera state
cameraControls.setOrbitPoint(0,0,0) -> dos not work with or without restore the orbit point
cameraControls.setPosition(position.x, position.y, position.z)
cameraControls.setTarget(target.x, target.y, target.z)
cameraControls.update()

cameraControls.getPosition() == position and 
cameraControls.getTarget() == target
but scene images is not same as when it was when saved the camera position
```

### Live example

https://yomotsu.github.io/camera-controls/examples/click-to-set-orbit-point.html

### Expected behavior

Camera should return to the same position

### Screenshots or Video

_No response_

### Device

Desktop

### OS

Windows

### Browser

Chrome

### LINKED ISSUE #424: "setOrbitPoint" will affect the "dollyToCursor" method cannot be maintained.
### Describe the bug

Hello, I am using camera-controls and want to achieve a function that can always rotate the camera around the origin. Therefore, I used setOrbitPoint. However, I found that the dollyToCursor set during the zoom process became invalid. I am not very familiar with this part of the code and I am not sure if it is a bug. I hope to get your help.

### To Reproduce

1. examples/orthographic.html demo
image
3. select dollyToCursor
4. drag
5. rotate
6. zoom

### Code

_No response_

### Live example

_No response_

### Expected behavior

maintain the effect of dollyToCursor.

### Screenshots or Video

https://github.com/yomotsu/camera-controls/assets/38150357/b7e667aa-888e-4333-9ad2-9fb30d361c38

### Device

Desktop

### OS

Windows

### Browser

Chrome

### LINKED ISSUE #491: Dolly to Cursor and focalOffset
### Describe the bug

Hi. thanks for this library. It helps a lot but i cant solve one problem.
I have OrthographicCamera and dollyToCursor working well but when i setOrbitPoint manually ( i want rotation around my part on scene) or make a movement with offset =>  after that when zooming with dollyToCursor it jumps by some delta. 
Video attached.

### To Reproduce

1. Set orbit point
2. Make dolly to cursor
3. See that it jumps by some delta

### Code

```js
this is creation of camera
const createOrthographicCamera = (name: string, position: Vector3, lookAt: Vector3) => {
            const camera = new OrthographicCamera(-aWidth, aWidth, aHeight, -aHeight, -10000, 10000);
            camera.position.copy(position);
            camera.lookAt(lookAt);
            camera.name = name;
            camera.zoom = 0.33
            return camera;
        };
//___________________________________
i set orbitPoint with help of this
cameraControls.setOrbitPoint(selectedPart.position.x, selectedPart.position.y, selectedPart.position.z)
//___________________________________
this is creation of cameraControls
this.mCameraControls = new CameraControls(camera,renderer.domElement);
this.mCameraControls.smoothTime = 0.15;
this.mCameraControls.minDistance = 1;
this.mCameraControls.maxDistance = 100000;
this.mCameraControls.minZoom = 1e-5;
this.mCameraControls.maxZoom = 1e5;
this.mCameraControls.dollyToCursor = true;
this.mCameraControls.maxPolarAngle = Math.PI;

this.mCameraControls.mouseButtons.right = CameraControls.ACTION.ROTATE
this.mCameraControls.mouseButtons.left = CameraControls.ACTION.NONE
this.mCameraControls.mouseButtons.middle = CameraControls.ACTION.TRUCK
this.mCameraControls.mouseButtons.wheel = CameraControls.ACTION.ZOOM
```

### Live example

_No response_

### Expected behavior

dolly to cursor dollies exactly to mouse location

### Screenshots or Video

https://github.com/yomotsu/camera-controls/assets/112622889/a48ae260-3191-406c-89d9-94aa8eee81b6

### Device

Desktop

### OS

Windows

### Browser

Chrome

CodyJasonBennett:
Fixes #303
Fixes #316
Fixes #424
Fixes #491

Applies and zeroes focal offsets into the control's spherical coordinates so features like zoom-to-cursor (e.g. `setOrbitPoint` and then interacting with an `OrthographicCamera`) and `setLookAt` work without interruption. I have not tested `setLookAt` as thoroughly, but expanded the fix to it after reading #303.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -3222,7 +3222,7 @@ export class CameraControls extends EventDispatcher {
 
 	// Applies and resets focal offsets to play nice with lookAt and zoom-to-cursor
 	// https://github.com/yomotsu/camera-controls/issues/491
-	_resetFocalOffsets() {
+	protected _resetFocalOffsets() {
 
 			_v3A.set( 0, 0, -1 ).applyQuaternion( this._camera.quaternion );
 			_v3A.multiplyScalar( this._spherical.radius );
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -723,6 +723,8 @@ export class CameraControls extends EventDispatcher {
 			if ( ! this._domElement ) return;
 			if ( ! this._enabled || this.mouseButtons.wheel === ACTION.NONE ) return;
 
+			this._resetFocalOffsets();
+
 			if (
 				this._interactiveArea.left !== 0 ||
 				this._interactiveArea.top !== 0 ||
@@ -842,6 +844,8 @@ export class CameraControls extends EventDispatcher {
 		const startDragging = ( event?: PointerEvent ): void => {
 
 			if ( ! this._enabled ) return;
+
+			this._resetFocalOffsets();
 
 			extractClientCoordFromEvent( this._activePointers, _v2 );
 
@@ -2034,6 +2038,8 @@ export class CameraControls extends EventDispatcher {
 		enableTransition: boolean = false,
 	): Promise<void> {
 
+		this._resetFocalOffsets();
+
 		this._isUserControllingRotate = false;
 		this._isUserControllingDolly = false;
 		this._isUserControllingTruck = false;
@@ -3213,6 +3219,19 @@ export class CameraControls extends EventDispatcher {
 	protected _addAllEventListeners( _domElement: HTMLElement ): void {}
 
 	protected _removeAllEventListeners(): void {}
+
+	// Applies and resets focal offsets to play nice with lookAt and zoom-to-cursor
+	// https://github.com/yomotsu/camera-controls/issues/491
+	_resetFocalOffsets() {
+
+			_v3A.set( 0, 0, -1 ).applyQuaternion( this._camera.quaternion );
+			_v3A.multiplyScalar( this._spherical.radius );
+			_v3A.add( this._camera.position );
+
+			this.setFocalOffset( 0, 0, 0, false );
+			this.moveTo( _v3A.x, _v3A.y, _v3A.z, false );
+
+	}
 
 	/**
 	 * backward compatible


yomotsu:
Sorry for the delay...
I believe I should check the changes carefully; however, I currently have limited time to do so.
A bit more time would be very helpful...

CodyJasonBennett:
No worries. These are sweeping changes, and people can work around it if they really want to today.

src/CameraControls.ts
```diff
@@ -723,6 +723,8 @@ export class CameraControls extends EventDispatcher {
 			if ( ! this._domElement ) return;
 			if ( ! this._enabled || this.mouseButtons.wheel === ACTION.NONE ) return;
 
+			this._resetFocalOffsets();
```
yomotsu, line 726: What about moving this line to the beginning of `_zoomInternal()` and `_dollyInternal()` with `if ( this.dollyToCursor )`?
Then wheel zoom/dolly will reset focalOffsets.

eg
```typescript
protected _dollyInternal = ( delta: number, x: number, y : number ): void => {

  if ( this.dollyToCursor ) this._resetFocalOffsets();
```

src/CameraControls.ts
```diff
@@ -843,6 +845,8 @@ export class CameraControls extends EventDispatcher {
 
 			if ( ! this._enabled ) return;
 
+			this._resetFocalOffsets();
```
yomotsu, line 848: I think mouse drag movement does not affect focalOffset. so, can we remove this?

src/CameraControls.ts
```diff
@@ -3214,6 +3220,19 @@ export class CameraControls extends EventDispatcher {
 
 	protected _removeAllEventListeners(): void {}
 
+	// Applies and resets focal offsets to play nice with lookAt and zoom-to-cursor
+	// https://github.com/yomotsu/camera-controls/issues/491
+	_resetFocalOffsets() {
```
yomotsu, line 3225: let's make this as a private method at this moment
```suggestion
	protected _resetFocalOffsets() {
```

src/CameraControls.ts
```diff
@@ -3214,6 +3220,19 @@ export class CameraControls extends EventDispatcher {
...
+
```
yomotsu, line 3226: Could we check the focalOffset value first?
If the value is 0, 0, 0, let's just return and do nothing.

src/CameraControls.ts
CodyJasonBennett: The idea is to call this whenever any input starts. Otherwise, we may interrupt transitions or animations. Is there a better place to put this?

Innovgame:
Thank you very much for fixing this issue. It indeed resolves #424, and I look forward to it being merged soon.

PR #529, opened 2024-07-21
feat(fitToSphere): add scale parameter to support scaling functionality

puxiao:
This is my current code:

```
const tempSphere = new Sphere()
CameraControls.createBoundingSphere(myMesh, tempSphere)
tempSphere.radius /= 0.8
cameraControls.fitToSphere(tempSphere, true)
```

I'm looking for a simpler code:

```
cameraControls.fitToSphere(myMesh, true, { scale: 0.8 })
```

So, I created this PR.

For the newly added parameter, I chose to use `{ scale: number }` instead of `scale:number` so that we can expand it to more parameters in the future if necessary.

### DIFF
--- a/examples/fit-to-bounding-sphere.html
+++ b/examples/fit-to-bounding-sphere.html
@@ -12,6 +12,8 @@
 <div class="info">
 	<a href="https://github.com/yomotsu/camera-controls">GitHub repo</a><br>
 	<button onclick="fit()">fit</button>
+	<button onclick="fit(0.5)">fit with scale 0.5</button>
+	<button onclick="fit(1.2)">fit with scale 1.2</button>
 	<button onclick="cameraControls.reset( true )">reset</button><br>
 </div>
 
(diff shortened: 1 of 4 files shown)


yomotsu:
Thank you for your PR with the document update.
What about simply scaling the given sphere before passing it to the method?

puxiao:
> Thank you for your PR with the document update. What about simply scaling the given sphere before passing it to the method?



```
fitToSphere( sphereOrMesh: _THREE.Sphere | _THREE.Object3D ,...)
```

`sphereOrMesh`：Sphere or Mesh

If it is a `sphere`, then there is no need to change it. You can indeed modify sphere first.

But if it's a `mesh`, it's a little inconvenient because you need to calculate the sphere first.

This is just my personal feeling. This PR only provides some simplifications, but it is not necessary.

yomotsu:
> But if it's a `mesh`

Thank you for explaining it in detail.

Similarly, another method called `fitToBox()` accepts extra sizes with the `padding` option, which is in WebGL units (such as `paddingTop: 1` for adding an extra `1` unit of space).
What you are looking for is in units? If so, padding{{ Direction }} would be more appropriate.
(Although scaling should still be usable for some cases.)

FYI: Actually, CameraControls provides a static method called `CameraControls.createBoundingSphere( `mesh` )`.

PR #544, opened 2024-10-16
chore: Improve readme.md

puxiao:


yomotsu:
Good catch!
Thanks for fixing.
Just merged

PR #549, opened 2024-12-16
Fix truckSpeed for orthographic camera

AlaricBaraou:
Previously truckSpeed parameter had no effect on the orthographicCamera

Tested in the ortho example, behavior is the same as basic.html ( Perspective )


yomotsu:
@AlaricBaraou 

Sorry for the long delay and thank you for noticing that.
That makes sense. Let me merge your PR.

PR #556, opened 2025-01-17
fix: Added missing call to _camera.updateMatrixWorld()

HTR1997:
Added missing call to _camera.updateMatrixWorld().

Currently the position of the camera is changed without this call. This results in issues where the rest of the scene believe the camera has not moved even though the rendered image makes it appear so.

https://github.com/user-attachments/assets/c4d091ec-50a8-4c40-907f-8e615b7eb88a

Above video shows on-hover highlighting. Notice that when the scene is panned that highlighting still occurs as though the tree is located in the center of the screen.


yomotsu:
@HTR1997 

Thank you for your PR, and sorry for the delay.
Aside from changing the camera offset, do we also need to update the matrix when rotating or moving its position?

I assume you are using a Raycaster, and I understand that we need to update the matrix when using it. However, I'm not sure if this should be done in cameraControls, because a matrix update may not be necessary for those who are not using a Raycaster. Also, updating the matrix will cause extra calculations. Therefore, I think it's better to update the matrix only when necessary on the user side.

HTR1997:
Thanks for getting back to me, no issue with delay.

I believe that this change can only be done within CameraControls. CameraControls sets its `_camera.matrix` on each call of `controls.update()`. Because `_camera.updateMatrixWorld()` is not called, the camera's `.matrixWorld` is not updated, which results in the rest of the scene not having the true location/orientation of the camera.

It seems that trying to call `camera.updateMatrixWorld()` outside of the update function does not update the location in time before its position is reset by `controls.update()`.

I will also note that the code *already* calls `_camera.updateMatrixWorld()`, and this PR changes where it is called, so it updates the location after the position is changed.

Regards, 
Harrison

yomotsu:
>  code already calls

Oops, You are right.
Sorry for my misunderstanding🙇‍♂️
Let me merge your PR.

PR #562, opened 2025-02-17
fix: Re-added _camera.updateMatrixWorld() as via _camera.matrix.compose()

HTR1997:
As a result of #556, there is some *wobbling* when rapidly rotating the scene (if using focalOffset). Re-adding the earlier call to `_camera.updateMatrixWorld` fixes this.

This PR directly calls `_camera.matrix.compose` rather than calling `updateMatrixWorld` to avoid unnecessary calculations.

Regards,
Harrison

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -2745,6 +2745,8 @@ export class CameraControls extends EventDispatcher {
 			! approxZero( this._focalOffset.z );
 		if ( affectOffset ) {
 
+			this._camera.matrix.compose( this._camera.position, this._camera.quaternion, this._camera.scale );
+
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for your PR.
I confirmed the wobbling problem:

https://github.com/user-attachments/assets/6e6617d3-6b81-4ba5-aa78-00e746588049

and your change will fix it.
Let me merge.

ilM4nzo:
I noticed that although #556 was merged into the dev branch, it doesn't seem to be included in the latest release v2.10.1

There's still some noticeable wobbling. Re-adding the previous call to _camera.updateMatrixWorld() or _camera.matrix.compose() appears to resolve this.

Just wanted to flag this in case it helps.

Best regards,
Robert

Issue #519, opened 2024-06-15
Trackpad Zoom/Scroll so slow in Orthographic Camera


arcanys-ivan:
### Describe the bug

When zooming using the laptop track pad, the behaviour is so slower compared to the mouse scroll

### To Reproduce

Just use Orthographic Camera
Use the laptop track pad instead of mouse
Zoom in and out

### Expected behavior

Should have the same smoothness and speed as the mouse scroll or zooming

babophett:
Try using smoothTime, dampingSmoothTime

seppestaes:
#461

Issue #461, opened 2023-10-04
Slow pinch-zoom when using orthographic camera


diredev:
### Describe the bug

Using the pinch gesture to zoom the scene works great when using a perspective camera.
If I use the gesture with an orthographic camera however, the camera moves very slowly, making the gesture pretty much unusable.

This is easily reproducable on mobile devices, I've tested it on both an iPad Pro 11 and on an Android phone.

### To Reproduce

1. Launch the "orthographic" camera example on a mobile device
2. Zoom using the pinch gesture

### Expected behavior

Should zoom at a reasonable speed even in orthographic mode.

### Device

Mobile

### OS

MacOS, Android

### Browser

Chrome, Safari

yomotsu:
Thank you for your report.
The problem has been reproduced.

I will take a look, but sorry, I don't have much time at this moment. It may take some time.

SrinivasPrabhu794:
Hi @yomotsu 

I have been facing the same issue of zoom in ortho camera. Can you point me in a direction where I can take a look so that I can take a dig at this issue?

Regards
Srinivas

SrinivasPrabhu794:
Hi @yomotsu 

Any help on this will be appreciated!

yomotsu:
sorry for the loooooong delay...
I checked the passed value when internal zoom happens:
https://github.com/yomotsu/camera-controls/blob/92ece6898f483e1f1844616a95ce0e839a48491f/src/CameraControls.ts#L3239

left is with a mouse wheel, and right is touch.
![image](https://github.com/yomotsu/camera-controls/assets/212837/ff05a3a5-da44-409a-93a1-fb07b67ba1ab)

seems the value is not much different on both sides.
but mouse wheel events happened more frequently if my investigation is correct...

What we could do is that:
Accelerate the delta a little bit, If the `internalZoom` is called from touch events.
(but not sure...
https://github.com/yomotsu/camera-controls/blob/92ece6898f483e1f1844616a95ce0e839a48491f/src/CameraControls.ts#L1165-L1200

SrinivasPrabhu794:
Thanks @yomotsu  :) Sorry for my late reply, ill take a look based on your inputs :) I appreciate your help on this!

SrinivasPrabhu794:
This helped me fix it for mobile devices. But the same pinch zoom slowness is observed when using the trackpad to pinch zoom on laptops in the orthographic mode. What solution would you suggest ?

ricardomatias:
[...]

PR #566, opened 2025-03-26
fix: trackpad pinch to zoom speed

pvalentim:
This aims to fix the issues #519 #461 where using the trackpad to zoom is very slow compared to the mouse wheel.

Fixed it by checking `ctrlKey` in the event and increasing the delta. This property is set by the browser to true when using multi touch gestures on the trackpad.

Since both issues reference using an orthographic camera and I'm using one myself in my project this needs to be tested using a perspective camera as well. I will try and find time to do it


yomotsu:
@pvalentim 

Sorry for the delay, and thank you for your PR.
I didn’t know that multi-touch gestures can be detected using event.ctrlKey, and should work correctly.

Would you mind adding a comment for event.ctrlKey?
For example: "Checks event.ctrlKey to detect multi-touch gestures on a trackpad."

yomotsu:
Thanks, Just merged!

PR #570, opened 2025-05-09
fix: incorrect camera direction on boundary zoom

MySongStan:
fix: Fixed an issue where when setting boundary and boundaryEncloseSComera to true, zooming back to the extreme position of the viewpoint with the mouse would result in incorrect looking direction of the camera, leading to errors in subsequent updates of the camera's cropping plane and causing some related calculation errors


yomotsu:
Thank you for the PR!
Before merging it, could you share the demo that’s causing the problem, just to be sure?

abernier:
maybe writing a repro test could also help here (once https://github.com/yomotsu/camera-controls/pull/592 is merged)

PR #571, opened 2025-05-09
docs: updating drei url

abernier:
drei doc has change, let's just update it :)


yomotsu:
Thanks for your PR as usual!
Just merged

PR #579, opened 2025-06-26
`lerpLookAt` normalizeRotations param

abernier:
add a `normalizeRotations` last param, to avoid normalizing rotation values

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -2127,6 +2127,7 @@ export class CameraControls extends EventDispatcher {
 	 * @param targetBZ
 	 * @param t
 	 * @param enableTransition
+	 * @param normalizeRotations
 	 * @category Methods
 	 */
 	lerpLookAt(
@@ -2136,6 +2137,7 @@ export class CameraControls extends EventDispatcher {
 		targetBX: number, targetBY: number, targetBZ: number,
 		t: number,
 		enableTransition: boolean = false,
+		normalizeRotations: boolean = true,
 	): Promise<void> {
 
 		this._isUserControllingRotate = false;
@@ -2164,7 +2166,9 @@ export class CameraControls extends EventDispatcher {
 			_sphericalA.theta  + deltaTheta  * t,
 		);
 
-		this.normalizeRotations();
+		if ( normalizeRotations ) {
+			this.normalizeRotations();
+		}
 
 		this._needsUpdate = true;
 


abernier:
I personally only need it inside `lerpLookAt` but indeed, a global `normalizeRotations()` method in user-land can make sense

yomotsu:
so, maybe, 
`cameraControls.lerpLookAt( ... )`
could be like
`cameraControls.normalizeRotations().lerpLookAt( ... )` if user needs.

would it be possbile to do that in this PR...?
(or I can also make a PR)

- Remove `normalizeRotations` from `setLookAt` and `lerpLookAt`.
- Make `normalizeRotations` return `this`, to make it chainable.
- replace `cameraControls.lerpLookAt` to `cameraControls.normalizeRotations().lerpLookAt` in examples

abernier:
you don't mind about introducing a breaking change, a v3?

if yes, this api seems gtm :)

.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: "20"
+          node-version-file: ".nvmrc"
```
abernier, line 22: i switched to .nvmrc to be able to use nvm (also added instructions into CONTRIBUTING.md)

CONTRIBUTING.md
```diff
@@ -1,6 +1,26 @@
...
+  $ nvm use
+  $ node -v # make sure your version satisfies package.json#engines.node
+  ```
+  nb: if you want this node version to be your default nvm's one: `nvm alias default node`
+- Install [npm](https://pnpm.io/installation#using-corepack), with:
+  ```sh
+  $ corepack enable
+  $ corepack prepare --activate # it reads "packageManager"
+  $ npm -v # make sure your version satisfies package.json#engines.npm
+  ```
```
abernier, line 17: proper instructions using the right version of node and npm using corepack

rollup.config.mjs
```diff
@@ -1,7 +1,9 @@
-import pkg from './package.json' assert { type: 'json' };
```
abernier, line 1: this was broken with my version of node, so i replace with:

```mjs
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const pkg = require('./package.json');
```

which seems safier

abernier (commented):
applied your instruction for normalizeRotations

+ chore things

abernier (commented):
@yomotsu can you review?

src/CameraControls.ts
```diff
@@ -2127,6 +2126,7 @@ export class CameraControls extends EventDispatcher {
 	 * @param targetBZ
 	 * @param t
 	 * @param enableTransition
+	 * @param normalizeRotations
```
yomotsu, line 2129: @abernier 
Is this arg still avalable?

abernier: oops yes removed

yomotsu:
Thanks for the changes.
Looks good to me.

It seems that moveTo() in examples also requires normalizeRotations(), but I’ll add that after merging this PR.

PR #591, opened 2025-07-28
JSON `.spherical` property

abernier:


src/types.ts
```diff
@@ -145,3 +145,32 @@ export type CameraControlsLerpState = {
 } | {
 	position: [number, number, number]
 } );
+
+export type CameraControlsState = {
```
abernier, line 149: a type for the state

PR #595, opened 2025-08-13
@abernier/camera-controls

abernier:


yomotsu:
Hi, @abernier
sorry for the long absence. (The situation is better now.)
I understand you put effort into this, and I really appreciate your support.
I'm rebooting now.

Issue #604, opened 2025-10-15
Zooming by pinching on a Macbook trackpad


npielawski:
### Describe the bug

Hello, and thanks for this nice library!

I am using an orthographic camera with camera controls wheel set to truck. I can pan using two fingers on my Macbook trackpad. If I pinch, it pans in the vertical direction, where I would expect it to Zoom. The touch gesture configuration doesn't alter the behavior.

Pinching inwards pans upwards, while pinching outwards pans downwards, which makes me believe the event is already captured but not linked to zooming.

### To Reproduce

1. pinch zoom using trackpad.

### Code

```js
<Canvas orthographic camera={{ zoom: 100 }}>
  <CameraControls
    mouseButtons={{
      wheel: ACTION.OFFSET
    }}
  />

  <mesh>
    <boxGeometry args={[1, 1, 1e-5]} />
    <meshStandardMaterial color="orange" />
  </mesh>
</Canvas>
```

### Expected behavior

It would be great if we could have the same behavior as Firefox's PDF viewer, where one pans with two fingers, while pinching zooms in and out.

### Device

Desktop

### OS

MacOS

yomotsu:
@npielawski 

Very sorry for the delay, and thanks for your research.
I also checked Figma’s zoom behavior, which accepts both trackpad pinch and Ctrl + wheel for zooming.

I think hard-coding Ctrl + wheel to zoom in makes sense.
Would it be okay to have your PR?

ShaMan123:
I am experiencing very slow zooming while pinching on the trackpad compared to mouse wheel.

PR #607, opened 2025-10-27
Fix: pinching now zooms in and out instead of scrolling up/down.

npielawski:
Fix for #604. Lets Apple trackpad users use the pinch gesture to zoom.

Zoom is triggered when ctrlKey is true. This happens both when pressing the CTRL key or when making a pinching gesture.

The zoom on Firefox has a different sensitivity than other browsers, so there was a condition (event.deltaMode == 0x01 [DOM_DELTA_LINE]) which helped set different sensitivities for different browsers. Firefox seems to be the only browser which sends 0x01 [DOM_DELTA_LINE], while the other ones send 0x00 [DOM_DELTA_PIXEL]. Source: https://stackoverflow.com/a/37474225

It seems that the sensitivity when pinching is *the same* across multiple browsers (I tested on Firefox, Chromium, and Safari), so I changed the condition to not change the sensitivity when pinching.

I tested the code by checking with the local examples, but also tested on a project of mine which uses React-Three-Drei, where one can truck in 2D and zoom. It worked well, and the same on all browsers (again, Firefox, Chromium and Safari).

I do not think documentation should be added, since this is an expected, intuitive and well-known behavior.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -767,11 +767,13 @@ export class CameraControls extends EventDispatcher {
 			// Ref: https://github.com/cedricpinson/osgjs/blob/00e5a7e9d9206c06fdde0436e1d62ab7cb5ce853/sources/osgViewer/input/source/InputSourceMouse.js#L89-L103
 			const deltaYFactor = isMac ? - 1 : - 3;
 			// Checks event.ctrlKey to detect multi-touch gestures on a trackpad.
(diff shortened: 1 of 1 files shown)


yomotsu:
Thanks for the PR. Looks good to me.

> I do not think documentation should be added

I agree. If any issues come up or users ask for clarification later, we can add a short note then.

Issue #611, opened 2025-11-03
Package is incompatible with `"moduleResolution": "nodenext"/"node16"`


helgee:
### Describe the bug

In Typescript projects that set `"moduleResolution": "nodenext"` or `"moduleResolution:":"node16" in their `tsconfig.json` this package cannot be imported.
Typescript reports the following error:
```
Property 'install' does not exist on type 'typeof import("[...]/node_modules/camera-controls/dist/index")'.ts(2339)
```

### To Reproduce

1. Set `"moduleResolution": "nodenext"` in `tsconfig.json`
2. `import CameraControls from 'camera-controls'`
3. Try to call `CameraControls.install(...)`

### Code

```js

```

### Expected behavior

Importing should work :-)

PR #612, opened 2025-11-03
fix: add `exports` and `type` field to package.json

helgee:
Fixes compatibility with strict ESM environments.

Refs: #611


src/index.ts
yomotsu: Do we really need this changes? I think If we're using TypeScript, we can omit the file extension.

yomotsu: Thank you for the explanation. I understand.
Since this is an npm library used in several projects, explicit file extensions should be added.

This might be wrong, but as a question. Would adding this to [rollup.config.mjs](https://github.com/yomotsu/camera-controls/blob/97fbe71dd9e278683e6677d64ccf2766fe81f5c2/rollup.config.mjs) solve the problem?

```diff
	output: [
+		{
+			format: 'es',
+			dir: 'dist',
+			banner: license,
+			indent: '\t',
+			exports: 'named',
+			preserveModules: true,
+			entryFileNames: '[name].js',
+			preserveModulesRoot: 'src',
+		},
		{
			format: 'es',
			file: pkg.module,
			banner: license,
			indent: '\t',
			exports: 'named',
		},
		{
			format: 'cjs',
			file: pkg.main,
			banner: license,
			indent: '\t',
			exports: 'named',
		}
```

helgee:
I have just pushed a third option that uses another Rollup plugin to achieve the same thing without changing the imports in the source code. I am fine with any of the three options but would probably go with #614 since it is the least intrusive and does not add maintenance overhead.

PR #623, opened 2025-12-07
Fix: Implement shortest arc theta interpolation in `update()`

WilliamVenner:
Hi, I noticed that when calling `setLookAt(..., enableTransition: true)`, it would sometimes interpolate a full circle when going from 360deg to 361deg, instead of interpolating by 1deg.

In this PR I have implemented shortest arc theta interpolation instead of the previously implemented linear interpolation.


WilliamVenner:
nevermind, looks like there's already a PR open for this #610 😁

PR #624, opened 2025-12-11
use camera's parent to transform target into world space

tormachjallwine:
Currently, if the camera is parented to a moving object, the orientation of the camera will change with it because the camera.lookAt function is used to orient the camera. The camera target should be transformed into world space first to allow for the camera to be parented to an object. This pull request performs this transformation.


yomotsu:
Nice finding. I'd like to merge this.
Thanks.

PR #635, opened 2026-07-22
fix: emit missed 'rest' before 'sleep' so transition listeners can't leak

wynged:
## The bug

`update()` dispatches `rest` only from the `else if ( updated )` branch — i.e. when the deltas fall within `restThreshold` **while the camera is still animating**. If the camera instead converges in the same frame a move starts, the next frame is not updated, so `update()` takes the `! updated && this._updatedLastTime` branch straight to `sleep` and skips `rest` entirely.

This is easiest to hit with `smoothTime = 0` (each move settles in the frame it starts), but any move small enough to settle within a single frame reaches it too.

Every transition-enabled call — `rotate` / `dolly` / `truck` / … , e.g. one per `pointermove` during a drag — registers a `rest` listener via `_createOnRestPromise()` that only detaches when `rest` fires. When `rest` is never emitted, those listeners, plus their promises and the closures they capture, accumulate **unbounded** for the life of the controls. A continuous drag under `smoothTime = 0` leaks one listener set per pointer-move.

We found this as the dominant retained-heap growth in a `three` app (heap snapshots showed one extra `Promise` + `onResolve` closure per pointer-move, surviving major GC).

## The fix

Emit the owed `rest` in the `sleep` branch, guarded on `! this._hasRested` so it is a strict no-op once `rest` has already fired the normal way. `sleep` already implies the camera has come to rest, so emitting the missing `rest` first is also more consistent with the documented event order (`rest` before `sleep`).

Because a `rest` listener may synchronously start a new transition, the `sleep` dispatch is also guarded on `! this._needsUpdate` — so `sleep` never fires in the same frame a `rest` handler kicked off new movement (the next frame emits `wake` as usual).

```ts
} else if ( ! updated && this._updatedLastTime ) {

    if ( ! this._hasRested ) {

        this._hasRested = true;
        this.dispatchEvent( { type: 'rest' } );

    }

    // A `rest` handler may have started a new transition; only sleep when
    // nothing is pending, so `sleep` never fires while the camera moves.
    if ( ! this._needsUpdate ) {

        this.dispatchEvent( { type: 'sleep' } );

    }

}
```

## Verification

- At normal `smoothTime`, `rest` fires from the existing branch and sets `_hasRested`, so the new guard is false → behavior is unchanged.
- With `smoothTime = 0`, a run of transition-enabled moves followed by a settle now drains the accumulated `rest` listeners to `0` instead of stranding one per move.
- A `rest` handler that starts a new transition no longer sees a spurious `sleep` while the camera is moving.

### DIFF
--- a/src/CameraControls.ts
+++ b/src/CameraControls.ts
@@ -2898,7 +2898,26 @@ export class CameraControls extends EventDispatcher {
 
 		} else if ( ! updated && this._updatedLastTime ) {
 
-			this.dispatchEvent( { type: 'sleep' } );
+			// A transition can stop without `update()` ever observing a settling
+			// frame — e.g. `smoothTime` 0, where each move converges in the same
+			// frame it starts — so the `rest` owed to `_createOnRestPromise()`
+			// listeners is never emitted and they accumulate unbounded. Emit it
+			// here before sleeping; the `_hasRested` guard keeps it a no-op once
+			// `rest` has already fired the usual way.
+			if ( ! this._hasRested ) {
+
+				this._hasRested = true;
+				this.dispatchEvent( { type: 'rest' } );
+
+			}
+
+			// A `rest` handler may have started a new transition; only sleep when
+			// nothing is pending, so `sleep` never fires while the camera moves.
+			if ( ! this._needsUpdate ) {
+
+				this.dispatchEvent( { type: 'sleep' } );
+
+			}
 
 		}
 


wynged:
any chance I could get a review on this?  would love to have it make it through rather than havinga local patch for this issue

beginnerJq:
Thanks for investigating this. We reproduced the same missing `rest` event in camera-controls 3.1.2 with the default `smoothTime`, when framing the same object again after the camera has fully settled. Our wrapper calls `fitToBox()` followed immediately by `rotateTo()`. The intermediate axis-aligned rotation is overwritten with the current angle, leaving a pending transition promise even though there is no final movement.

Here is a standalone Node.js reproduction using Three.js 0.185.1:

```js
import * as THREE from 'three';
import CameraControls from 'camera-controls';

globalThis.DOMRect ??= class DOMRect {};
CameraControls.install({ THREE });
const camera = new THREE.PerspectiveCamera(50, 1, 0.1, 1000);
camera.position.set(0, 0, 30);
const controls = new CameraControls(camera);
const box = new THREE.Box3(
  new THREE.Vector3(-5, -5, -5),
  new THREE.Vector3(5, 5, 5),
);
const fly = () => Promise.all([
  controls.fitToBox(box, true),
  controls.rotateTo(0, Math.PI / 4, true),
]);
async function tick() {
  for (let i = 0; i < 1000; i++) {
    controls.update(1 / 60);
    await Promise.resolve();
  }
}

let first = false;
fly().then(() => { first = true; });
await tick();

const events = [];
for (const type of ['wake', 'rest', 'sleep']) {
  controls.addEventListener(type, () => events.push(type));
}
let second = false;
fly().then(() => { second = true; });
await tick();
console.log({ first, second, events });
// { first: true, second: false, events: ['wake', 'sleep'] }

controls.rotate(0.1, 0, true);
await tick();
console.log({ secondAfterMovement: second });
// { secondAfterMovement: true }
```

The second promise remains pending despite 1,000 further `update()` calls, and moving the camera releases it. This seems like a useful additional regression case for this PR: repeated framing after full sleep, with a transition target overwritten back to the current position. We currently work around it downstream by resolving on either `rest` or `sleep` and removing both listeners on completion; we have not yet tested this PR branch itself.

